PR #30, opened 2020-05-12
import name/homepage as keyword

Mic92:
otherwise elasticsearch will performs stemming which removes important
information:

Example:

Name: "nixpkgs-review"
Elasticsearch with 'text' mapping: ["nixpkgs", "review"]

At the moment searching for `*nixpkgs-review*` will return no results.

### DIFF
--- a/scripts/import-channels-into-elasticsearch
+++ b/scripts/import-channels-into-elasticsearch
@@ -155,7 +155,7 @@ def recreate_index(es, channel):
             mappings=dict(
                 properties=dict(
                     attr_name=dict(type="keyword"),
-                    name=dict(type="text"),
+                    name=dict(type="keyword"),
                     version=dict(type="text"),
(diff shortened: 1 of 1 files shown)


Mic92:
This change might be not required. I found out that match queries, still find the information fine.

Checkout this prototype I built: https://gist.github.com/Mic92/6d153e5d1d21d2ce3bb4da4c03b29082

garbas:
I think this should be keyword too.

Anyway right now are only doing wildcard query on `attr_name` and `option_name`. This was done just to reach the [parity goal](https://github.com/NixOS/nixos-search/projects/1) first before focusing on things such as ranking, aggregation, custom tokenizer, ...

Another thing I would like to mention is that if you plan to use elasticsearch from other tools, then I don't guarantee *schema* will not change. On the contrary we will configure as often as possible to improve the functionality of the frontend.

PR #54, opened 2020-05-31
use one index per channel

### LINKED ISSUE #23: reduce shards in elastic search
right now we create 2 indexes per channel. one for packages and one for options. the data from one evaluation should be in one index.

garbas:
fixes #23


garbas:
Looks like the search was screwed up a little.

PR #97, opened 2020-06-12
Fix title of the app before it loads

### LINKED ISSUE #93: [review] Page title before loading the app is misleading
*As part of the review for https://github.com/NixOS/nixos-homepage/pull/454*

* * *

```
<title>Elm hotloading dev environment</title>
```

We probably want to set it to "NixOS Search", and *if desired*, probably means some non-Elm JS, add *— Loading...* or something similar.

Main issue with keeping this title is that web search results may show a completely bogus page name!!

garbas:
and also improve on window titles for other pages

fixes #93

### DIFF
--- a/src/Main.elm
+++ b/src/Main.elm
@@ -252,39 +252,60 @@ update msg model =
 -- VIEW
 
 
-view : Model -> Html Msg
+view :
+    Model
+    ->
+        { title : String
+        , body : List (Html Msg)
+        }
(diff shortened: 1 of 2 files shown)


samueldr:
One issue with adding `Loading...`  to the non-js'd version of the page is that our entries in search results may show `NixOS Search - Loading...` depending how much JS the crawler executes.

PR #99, opened 2020-06-12
handle 10000 package limit in the UI correctly

### LINKED ISSUE #94: [review] Large result sets pagination trouble
*As part of the review for https://github.com/NixOS/nixos-homepage/pull/454*

* * *

I am unsure what is going on, but bear with what I am observing:

https://search.nixos.org/packages?query=haskell&from=0&size=15&channel=unstable

 * Search for `haskell`
 * Observe results are an oddly round *0-15 of 10000*.
 * Try to access last page
 * Observe  
   image
 * Observe URL referencing `&from=9990`

I guess the app will need to *somehow* act in a less error-y way for these cases. First off, I'd say that if the size can be understood as *indefinite*, we should show *of too many results* or something like that. Then, disable the "Last" page navigation, and when going over the limit (is it detectable?) drop them into a page saying they should search with more precise terms, or something of the like.

garbas:
fixes #94


src/Search.elm
```diff
@@ -402,8 +402,13 @@ view path title model viewSuccess outMsg =
                                                 model.from + model.size
                                             )
                                         ++ " of "
-                                        ++ String.fromInt result.hits.total.value
-                                        ++ "."
+                                        ++ (if result.hits.total.value == 10000 then
+                                                "more then 10000. You have reached the maximum of search results we can show."
```
samueldr, line 406:
```suggestion
                                                "more than 10000. You have reached the maximum of search results we can show."
```

`more then` → `more than`.

I don't like "You have reached the maximum of search results we can show", though not sure what to replace it with.

First off, we need to invite the user to make a more precise search, only being sorry isn't helpful. It's also an implementation detail to say that it's the "maximum we can show".

Maybe something along the lines of "*more than 10000 results, please provide more precise search terms.*"

If we're informal in the app, maybe something like: "*more than 10000 results, that's way too many to be useful. Help us help you and make a more precise search!*"

src/Search.elm
```diff
@@ -511,7 +516,7 @@ viewPager outMsg model result path =
             ]
         , li
             [ classList
-                [ ( "disabled", model.from + model.size >= result.hits.total.value )
+                [ ( "disabled", result.hits.total.value == 10000 || model.from + model.size >= result.hits.total.value )
```
samueldr, line 519: It's not really disabled, it can still be used by clicking on it, and it goes to the proper last page.

src/Search.elm
garbas: I'll go for the boring formal version.

Issue #157, opened 2020-08-25
Contributing on MacOS


hmaurer:
Hello!

Awesome work on the new search; I noticed it being a lot snappier and headed over to this repo to figure out what changed :-) I've been trying to compile the project from source on MacOS but I run into an issue during the yarn installation step:

```
nixos-search % nix-shell --run "yarn dev"
these 2 derivations will be built:
  /nix/store/jkg46ddn2q9xn7rhi39ba37lx8fw3ibn-nixos-search-modules-1.0.0.drv
  /nix/store/6vrdv2am9bpdh8j76hlrx9i33vax7crh-nixos-search-yarn-1.0.0.drv
building '/nix/store/jkg46ddn2q9xn7rhi39ba37lx8fw3ibn-nixos-search-modules-1.0.0.drv'...
configuring
building
'/private/tmp/nix-build-nixos-search-modules-1.0.0.drv-0/yarn_home/.node-gyp/12.18.3/include' -> '/nix/store/h8gi5368bw7rxcs153i55xb2smak5470-nodejs-12.18.3/include'
yarn config v1.22.4
success Set "yarn-offline-mirror" to "/nix/store/gqi0wz5ipx7c39y7l8hyscwkxxf9a62q-offline".
✨  Done in 0.03s.
yarn install v1.22.4
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[...]

PR #167, opened 2020-08-28
switch from sass to less (css)

hmaurer:
This PR switches the CSS preprocessor from https://sass-lang.com/ to http://lesscss.org/. This resolves an issue with `node-sass` on OS X (https://github.com/NixOS/nixos-search/issues/157) and will facilitate integration with  https://github.com/NixOS/nixos-homepage/pull/506.

I'd appreciate a careful review of the Nix side of these changes. I removed a chunk of code from `default.nix` which appeared to be customising `node-sass`, but I don't know if something similar is necessary for `less`. Everything appears to work as expected, though.

PS: question to @garbas (or another contributor): how do you generate `yarn.lock` after making changes to `package.json`? Yarn does not seem to have a flag to generate a lock file only, so I've been running `yarn install && rm -r node_modules`. Is there another way?

If you don't already have a better solution in place we could add a command to the nix shell using `npm --package-lock-only` and https://github.com/imsnif/synp, c.f. https://stackoverflow.com/a/58609139/2362799


garbas:
@hmaurer we kinda have to use yarn since only yarn2nix exists. there is also npm2nix but that requires us to generate nix file and not just reading the lock file. We can switch once situation changes there.

There is also `yarn generate-lock-entry`, but `yarn install` also works.

I tested this locally and works fine!

hmaurer:
@garbas ah! just to be clear: I wasn't suggesting switching to NPM. `yarn generate-lock-entry` does not, as far as I know, generate a `yarn.lock`. What I was suggesting is the solution described in the stackoverflow answer I linked: use `npm install --package-lock-only` to generate a `package-lock.json` file, then convert it to `yarn.lock` via `synp`. But perhaps that's overthinking it; `yarn install && rm -r node_modules` works.

PR #168, opened 2020-08-28
rework ranking queries

### LINKED ISSUE #156: Search seems to be OR-based now reducing relevancy for multiple terms
Results are far less relevant with OR-based search:

 * minifier: 32 results
 * css: 535 results
 * css minifier: 559 results

I have to go through all "css" and "minifier" results when I'm searching for what should have been "css AND minifier".

Is that what 91527f1deb18e3f74b6e52c50cadc08b4366e4f2 changed?

garbas:
fixes #156

---
This change is [Reviewable](https://reviewable.io/reviews/nixos/nixos-search/168)

### DIFF
--- a/src/Page/Options.elm
+++ b/src/Page/Options.elm
@@ -279,132 +279,19 @@ makeRequest :
     -> Int
     -> Search.Sort
     -> Cmd Msg
-makeRequest options channel queryRaw from size sort =
-    let
-        query =
-            queryRaw
-                |> String.trim
-
-        delimiters =
-            Maybe.withDefault Regex.never (Regex.fromString "[. ]")
-
-        should_match boost_base =
-            List.indexedMap
-                (\i ( field, boost ) ->
-                    [ ( "match"
-                      , Json.Encode.object
-                            [ ( field
-                              , Json.Encode.object
-                                    [ ( "query", Json.Encode.string query )
-                                    , ( "boost", Json.Encode.float <| boost_base * boost )
-                                    , ( "analyzer", Json.Encode.string "whitespace" )
-                                    , ( "fuzziness", Json.Encode.string "1" )
-                                    , ( "_name"
-                                      , Json.Encode.string <|
-                                            "should_match_"
-                                                ++ String.fromInt (i + 1)
-                                      )
-                                    ]
-                              )
-                            ]
-                      )
-                    ]
-                )
-                [ ( "option_name", 1 )
-                , ( "option_name_query", 1 )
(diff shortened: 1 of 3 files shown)


samueldr:
It looks like it may be filtering results way better, for some of the queries. `css minifier` decreases from 559 to 192 results.

Though, at the same time, there are still high ranking results that I doubt should be present. Right after the perl-based CSS minifier results, there are JavaScript minifiers, nothing matching CSS AFAICT.

 * https://5f49651ee28abe00aa84b74a--nixos-search.netlify.app/packages?show=perl528Packages.JavaScriptMinifierXS&query=css%20minifier&from=0&size=15&sort=relevance&channel=unstable

I don't understand what the implementation changes here are, considering my lack of expertise with Elasticsearch. So I can't say for sure whether this is going in the right direction, though observing the effects it seems it does.

Is it possible to see why a result was matched in the result? Is this what `matched_queries` represent? And then it's ranked (only?) by the amount of matched queries? I see it goes from 20 for the first two, while going to 9 from then on. I can see the same names under `should` in the query.

garbas:
> It looks like it may be filtering results way better, for some of the queries. `css minifier` decreases from 559 to 192 results.
> 
> Though, at the same time, there are still high ranking results that I doubt should be present. Right after the perl-based CSS minifier results, there are JavaScript minifiers, nothing matching CSS AFAICT.
> 
>     * https://5f49651ee28abe00aa84b74a--nixos-search.netlify.app/packages?show=perl528Packages.JavaScriptMinifierXS&query=css%20minifier&from=0&size=15&sort=relevance&channel=unstable
> 
> 
> I don't understand what the implementation changes here are, considering my lack of expertise with Elasticsearch. So I can't say for sure whether this is going in the right direction, though observing the effects it seems it does.
> 

I'm thinking of merging this since it immediately improves things, but trying to rethink and improve further.

> Is it possible to see why a result was matched in the result? Is this what `matched_queries` represent? And then it's ranked (only?) by the amount of matched queries? I see it goes from 20 for the first two, while going to 9 from then on. I can see the same names under `should` in the query.

in both Packages.elm and Options.elm you can look for `-- DEBUG:` and if you comment out those lines it will print which queries are matched. Also the matching score is then displayed for each result. This is what I usually uncomment when I'm working on figuring out the relevancy.

But you don't have to worry with Elm, you just need to type JSON as described in https://www.elastic.co/guide/en/elasticsearch/reference/7.2/query-dsl.html . then you simply issue a curl request.

PR #172, opened 2020-08-29
using whitespace analyzer and not the default one

### LINKED ISSUE #171: Results do not contain the search query at all
Hi,

I was searching for all packages related to chinese fonts or language packs, and usually they contain "zh-cn" in their name in many distributions (ex: https://aur.archlinux.org/packages/?O=0&K=zh-cn).

However, this exact same search query (https://search.nixos.org/packages?query=zh-cn&from=0&size=15&sort=relevance&channel=20.03), leads to lots of results, but they **all** appear to be unrelated, and do not even contain "zh-cn" anywhere (just Ctrl-f it).

image

I suppose that the search engine may try to be smart about the query, and searches for related terms, displaying those.

Expected: if there are no exact match, it would be nice to know about it at least.

garbas:
fixes #171

---
This change is [Reviewable](https://reviewable.io/reviews/nixos/nixos-search/172)

### DIFF
--- a/src/Search.elm
+++ b/src/Search.elm
@@ -751,6 +751,7 @@ search_fields query fields =
                         [ ( "type", Json.Encode.string "most_fields" )
                         , ( "query", Json.Encode.string queryWord )
                         , ( "fuzziness", Json.Encode.int <| String.length queryWord // 5 )
+                        , ( "analyzer", Json.Encode.string "whitespace" )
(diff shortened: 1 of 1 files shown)


Jiehong (approved):
I just tried, and it does seem to fix #171 ! thanks!

Jiehong (approved):

garbas:
@Jiehong thank you for reporting.

PR #195, opened 2020-09-17
By mistake the default analyzer was used for query

### LINKED ISSUE #187: Regression for searches with placeholders like `<name>`
#149 was closed, but I believe the "fixed" behavior is broken too: instead of returning zero results, it returns way too many, rendering the search pretty useless.

It seems most people do prefix searches in search.nixos.org. So when searching for `nginx.<name>`, getting `services.grafana.database.name` as a [result](https://search.nixos.org/options?show=services.grafana.database.name&query=nginx.%3Cname%3E&from=30&size=30&sort=relevance&channel=unstable) makes very little sense.

I suggest treating `<[^\s>]+>` as special in the tokenizer, instead of making `<` and `>` behave like whitespace. IMHO, the `<foo>` convention (is it really a convention?) for placeholders is widespread enough to warrant special-casing it in the search engine.

### LINKED ISSUE #191: Package search is very confusing now
Specifically, searching for a specific package, especially if it has non-alphanumerics in the name, tends to return lots of unrelated results.

For example:
- A search for `munin` returns one result, munin itself
- A search for `contrib` returns 103 results. Not all of them have "contrib" in the name but the ones that don't appear to have it in the description, so that's ok (although a "search in description y/n" option would be nice)
- A search for `munin contrib` returns no results; presumably it's searching for (`munin` and `contrib`), which is the behaviour I expect.
- A search for `munin-contrib` (the actual package I was looking for) returns...40 results? It's clearly not the same as (`munin` or `contrib`), but it's not (`munin` and `contrib`) either, or `munin-contrib` as a single string (which is what I actually wanted)

This appears to have been introduced quite recently, possibly at the same time as the new "fuzzy" search.

garbas:
At the same time i'm removing all of the fuzziness.

fixes #191 
fixes #187

---
This change is [Reviewable](https://reviewable.io/reviews/nixos/nixos-search/195)

### DIFF
--- a/src/Search.elm
+++ b/src/Search.elm
@@ -767,7 +767,8 @@ search_fields baseScore queryWords fields =
                   , Json.Encode.object
                         [ ( "type", Json.Encode.string "bool_prefix" )
                         , ( "query", Json.Encode.string queryWord )
-                        , ( "fuzziness", Json.Encode.int <| String.length queryWord // 5 )
+                        , ( "analyzer", Json.Encode.string "lowercase" )
+                        , ( "auto_generate_synonyms_phrase_query", Json.Encode.bool False )
                         , ( "prefix_length", Json.Encode.int 3 )
                         , ( "operator", Json.Encode.string "or" )
                         , ( "_name"


ToxicFrog:
Thank you for the quick fix!

PR #214, opened 2020-10-27
Percent encode query

### LINKED ISSUE #200: URL query encoding of spaces is broken
Spaces in the query part of an URL should be encoded as `+` ([source](https://stackoverflow.com/a/29948396)).
`nixos-search` uses `%20` instead.
This results in url https://search.nixos.org/options?query=a+b starting a query for `a+b` instead of `a b`, thereby breaking external tools like browser search plugins.

turboMaCk:
resolves #200

### This PR has 2 commits:

1. Fixes the issue itself
1. Unification of `createUrl` function in search with `routeToString` to ensure compatibility of urls

### Showcase

nix-search

### DIFF
--- a/elm.json
+++ b/elm.json
@@ -1,39 +1,40 @@
 {
-  "type": "application",
-  "source-directories": [
-    "src"
-  ],
-  "elm-version": "0.19.1",
-  "dependencies": {
-    "direct": {
-      "NoRedInk/elm-json-decode-pipeline": "1.0.0",
-      "elm/browser": "1.0.2",
-      "elm/core": "1.0.4",
-      "elm/html": "1.0.0",
-      "elm/http": "2.0.0",
-      "elm/json": "1.1.3",
-      "elm/regex": "1.0.0",
-      "elm/url": "1.0.0",
-      "hecrj/html-parser": "2.3.4",
-      "krisajenkins/remotedata": "6.0.1",
-      "truqu/elm-base64": "2.0.4"
+    "type": "application",
+    "source-directories": [
+        "src"
+    ],
+    "elm-version": "0.19.1",
+    "dependencies": {
+        "direct": {
+            "NoRedInk/elm-json-decode-pipeline": "1.0.0",
+            "elm/browser": "1.0.2",
+            "elm/core": "1.0.4",
+            "elm/html": "1.0.0",
+            "elm/http": "2.0.0",
+            "elm/json": "1.1.3",
+            "elm/regex": "1.0.0",
+            "elm/url": "1.0.0",
+            "elm-community/basics-extra": "4.1.0",
+            "hecrj/html-parser": "2.3.4",
+            "krisajenkins/remotedata": "6.0.1",
+            "truqu/elm-base64": "2.0.4"
+        },
+        "indirect": {
+            "elm/bytes": "1.0.8",
+            "elm/file": "1.0.5",
+            "elm/parser": "1.1.0",
+            "elm/time": "1.0.0",
+            "elm/virtual-dom": "1.0.2",
+            "rtfeldman/elm-hex": "1.0.0"
+        }
     },
-    "indirect": {
-      "elm/bytes": "1.0.8",
-      "elm/file": "1.0.5",
-      "elm/parser": "1.1.0",
-      "elm/time": "1.0.0",
-      "elm/virtual-dom": "1.0.2",
-      "rtfeldman/elm-hex": "1.0.0"
+    "test-dependencies": {
+        "direct": {
+            "elm-explorations/test": "1.2.2"
+        },
+        "indirect": {
+            "elm/random": "1.0.0",
+            "elm/svg": "1.0.1"
+        }
     }
-  },
-  "test-dependencies": {
-    "direct": {
-      "elm-explorations/test": "1.2.2"
-    },
-    "indirect": {
-      "elm/random": "1.0.0",
-      "elm/svg": "1.0.1"
-    }
-  }
 }
--- a/elm.json
+++ b/elm.json
@@ -1,40 +1,39 @@
 {
-    "type": "application",
-    "source-directories": [
-        "src"
-    ],
-    "elm-version": "0.19.1",
-    "dependencies": {
-        "direct": {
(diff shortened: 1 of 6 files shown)


elm.json
```diff
@@ -1,39 +1,40 @@
...
+        "direct": {
+            "NoRedInk/elm-json-decode-pipeline": "1.0.0",
+            "elm/browser": "1.0.2",
+            "elm/core": "1.0.4",
+            "elm/html": "1.0.0",
+            "elm/http": "2.0.0",
+            "elm/json": "1.1.3",
+            "elm/regex": "1.0.0",
+            "elm/url": "1.0.0",
+            "elm-community/basics-extra": "4.1.0",
```
turboMaCk, line 17: this is the package which was added. Probably this file was added by hand before so `elm install` changed it. Let me know if this should be reverted.

Also I only use `uncurry` function from the package - we don't necessary need it - let me know if you want me to remove it. Anyway it provides rather standard lambda calculus style combinatorics missing in elm's Basics module.

src/Route.elm
```diff
@@ -19,34 +21,35 @@ type Route
     | Options (Maybe String) (Maybe String) (Maybe String) (Maybe Int) (Maybe Int) (Maybe String)
 
 
+{-| Fixes issue with elm/url not properly escaping string
+-}
+queryString : String -> Query.Parser (Maybe String)
+queryString =
+    Query.map (Maybe.andThen Url.percentDecode) << Query.string
+
+
```
turboMaCk, line 30: Since [Domen's PRs](https://github.com/elm/url/pull/34) are still not merged we need to define fixed version ourselves.

src/Route.elm
```diff
@@ -19,34 +21,35 @@ type Route
...
-            Home
-            Url.Parser.top
-        , Url.Parser.map
-            NotFound
-            (Url.Parser.s "not-found")
-        , Url.Parser.map
-            Packages
+        [ Url.Parser.map Home Url.Parser.top
+        , Url.Parser.map NotFound (Url.Parser.s "not-found")
+        , Url.Parser.map Packages
```
turboMaCk, line 36: I took a liberty and reformatted these parts a bit. This is all elm-format compatible.

src/Route.elm
```diff
@@ -79,41 +82,64 @@ fromUrl url =
...
+routeToString =
+    uncurry Builder.absolute << routeToPieces
 
 
-routeToPieces : Route -> ( List String, List (Maybe String) )
+{-| Fixes issue with elm/url not properly escaping string
+-}
+builderString : String -> String -> QueryParameter
+builderString name =
+    Builder.string name << Url.percentEncode
```
turboMaCk, line 93: Since [Domen's PRs](https://github.com/elm/url/pull/34) are still not merged we need to define fixed version ourselves.

src/Route.elm
```diff
@@ -79,41 +82,64 @@ fromUrl url =
...
 
 routeToString : Route -> String
-routeToString page =
-    let
-        ( path, query ) =
-            routeToPieces page
-    in
-    "/" ++ String.join "/" path ++ "?" ++ String.join "&" (List.filterMap Basics.identity query)
+routeToString =
+    uncurry Builder.absolute << routeToPieces
```
turboMaCk, line 86: this is the only use of uncurry from basics-extra

src/Search.elm
```diff
@@ -275,39 +287,20 @@ update path navKey msg model =
...
-                    (\x ->
-                        [ Url.Builder.string "show" x
-                        ]
-                    )
-                |> Maybe.withDefault []
-            )
-        |> Url.Builder.absolute [ path ]
+createUrl toRoute channel query show from size sort =
+    toRoute (Just channel) query show (Just from) (Just size) (Just <| toSortId sort)
+        |> Route.routeToString
```
turboMaCk, line 303: unification of URL building logic so we can keep fixes in single place.

src/Search.elm
```diff
@@ -320,6 +313,7 @@ type Channel
     | Release_20_03
     | Release_20_09
 
+
```
turboMaCk, line 316: elm format.

elm.json
garbas: I tend to avoid adding dependencies (if possible)

Since we are only using one function would it be too much to just copy/paste it (with a comment from where we pasted it)? otherwise we depend on whole basics-extra stuff. 

I find this to work better when maintaining code and also it is much easier to port to new elm versions. I don't have the exact rule but adding a dependency for only one function is a bit of an overkill.

src/Page/Options.elm
```diff
@@ -102,7 +103,7 @@ update navKey msg model =
             let
                 ( newModel, newCmd ) =
                     Search.update
-                        "options"
+                        Route.Options
```
garbas, line 106: :heart:

src/Route.elm
garbas: :100:

elm.json
turboMaCk: > I tend to avoid adding dependencies (if possible)

yeah I'm the same but I learned to make an exception with elm-community stuff. Anyway lets keep this PR focused - I'll remove the dependency and do the uncurryng manually.

src/Route.elm
```diff
@@ -1,11 +1,13 @@
...
 
+import Basics.Extra exposing (uncurry)
 import Browser.Navigation
 import Html
 import Html.Attributes
 import Url
+import Url.Builder as Builder exposing (QueryParameter)
 import Url.Parser exposing ((<?>))
-import Url.Parser.Query
+import Url.Parser.Query as Query
```
garbas, line 9: I like to avoid using aliases and prefer to use absolute names.

I know it makes this bigger, but I find it much nicer when reading the code. Especially after some time, since I don't have to look at the header where the function is coming from.

If you don't have strong feelings about this I would like to revert this.

src/Route.elm
garbas: :100:

src/Search.elm
garbas: :1st_place_medal:

garbas (changes requested):
I've added the feedback. I'm closing the PR since @turboMaCk will push the branch to repo itself so we get nice previews.

PR #215, opened 2020-10-27
Support search engine escaping of spaces (+)

### LINKED ISSUE #200: URL query encoding of spaces is broken

turboMaCk:
resolves #200

Interpret `+` in seaerch query string as space.

**There is a limitation imposed by elm URL parser** we can't access get the raw urls string even when using `Url.Parser.Query.custom` -- the string argument is already decoded. this means we can't tell if the `+` comes from url or was decoded from percent string (`%2B`). All `+` are replaced with space.

### Showcase

updated-search

previous PR: https://github.com/NixOS/nixos-search/pull/214

### DIFF
--- a/src/Route.elm
+++ b/src/Route.elm
@@ -1,6 +1,5 @@
 module Route exposing (Route(..), fromUrl, href, replaceUrl, routeToString)
 
-import Basics.Extra exposing (uncurry)
 import Browser.Navigation
 import Html
 import Html.Attributes
@@ -83,7 +82,7 @@ fromUrl url =
 
 routeToString : Route -> String
 routeToString =
-    uncurry Builder.absolute << routeToPieces
+    (\( path, query ) -> Builder.absolute path query) << routeToPieces
 
 
 {-| Fixes issue with elm/url not properly escaping string
--- a/src/Route.elm
+++ b/src/Route.elm
@@ -9,6 +9,17 @@ import Url.Parser exposing ((<?>))
 import Url.Parser.Query
 
 
+{-| This is type safe wrapper for working with search queries in url
+-}
+type SearchQuery
+    = SearchQuery String
+
+
+unSearchQuery : SearchQuery -> String
+unSearchQuery (SearchQuery str) =
+    str
+
+
 
 -- ROUTING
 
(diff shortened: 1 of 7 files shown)


turboMaCk:
@garbas Feel free to check last commit https://github.com/NixOS/nixos-search/pull/215/commits/95de1b73458b8db38a280b6ab4ae1f99586e6650 but also be aware of [the limitations](https://github.com/NixOS/nixos-search/pull/215#issue-510933385)

src/Route.elm
```diff
@@ -19,30 +31,32 @@ type Route
...
-            (Url.Parser.s "not-found")
-        , Url.Parser.map
-            Packages
+        [ Url.Parser.map Home Url.Parser.top
+        , Url.Parser.map NotFound (Url.Parser.s "not-found")
+        , Url.Parser.map Packages
             (Url.Parser.s "packages"
                 <?> Url.Parser.Query.string "channel"
-                <?> Url.Parser.Query.string "query"
+                <?> Url.Parser.Query.map (Maybe.map unSearchQuery) (querySearchString "query")
```
turboMaCk, line 50: for now unwrapping inner string directly here so type of routes doesn't need to change.

garbas:
This now works when we load the app with the url that has `+` in the query parameter of the query string.
But if we input `+` into the input box it gets replaced by space.

turboMaCk:
yes exactly. Anyway I'm afraid there is no way around the 2nd problem because even with `custom` function from elm url parser the strings one work with are already decoded which makes it impossible to tell the which plus is plus and which one is space:/

garbas:
I guess we will have to resort to ports and implement our own wrapper around `Url.Url` where we would take values from `window.location` via ports into the account. That should allow us to distinguish between "space/+" and "+"

turboMaCk:
I would first like to check where does the escaping happens. If it's in Browser.Navigation or within Url.Parser. If it's the later there would also be option to "patch this on just a parser level" by accessing the raw string. The only kernel function of Url parser seems to be binding to decode/encodeURIComponent. https://github.com/elm/url/blob/master/src/Elm/Kernel/Url. If we'll be able to get undecoded data from navigation object in elm we might consider either:

1. replacing url Parser with implementation that gives us access to the raw string
2. using navigation object as a workaround.

I'll let you know once I know more.

turboMaCk:
@garbas See the last commit - it implements pure elm solution. In theory it would be possible to just wrap whole elm/url `Parser` and embed it into larger custom Parser structure to build new nice API in which it would be possible to define this function as part of `Parser` type. Anyway that would be a lot of code. Current approach is an attempt for pragmatic compromise - not a complete hack but not a full blown custom url parser library either. Let me know what you think.

garbas (approved):
Works great. And I like the solution.

PR #216, opened 2020-10-27
Remove cleanUrl and fragment hack

turboMaCk:
I just removed the `cleanUrl` function and updated all the places I found where `#disabled` fragment was used. I think this is enough to remove the hack. I tested the behavior there are no refreshes going on.

### DIFF
--- a/src/Main.elm
+++ b/src/Main.elm
@@ -188,7 +188,7 @@ changeRouteTo currentModel url =
         Just route ->
             let
                 model =
-                    { currentModel | route = Debug.log "route" route }
+                    { currentModel | route = route }
             in
             case route of
                 Route.NotFound ->
--- a/src/Main.elm
+++ b/src/Main.elm
@@ -159,6 +159,26 @@ changeRouteTo :
     -> Url.Url
     -> ( Model, Cmd Msg )
 changeRouteTo currentModel url =
+    let
+        attempteQuery (( newModel, cmd ) as pair) =
+            case ( currentModel.route, newModel.route ) of
+                ( Route.Packages c1 q1 _ f1 s1 srt1, Route.Packages c2 q2 _ f2 s2 srt2 ) ->
+                    if c1 /= c2 || q1 /= q2 || f1 /= f2 || srt1 /= srt2 then
+                        submitQuery newModel ( newModel, cmd )
+
+                    else
+                        pair
+
+                ( Route.Options c1 q1 _ f1 s1 srt1, Route.Options c2 q2 _ f2 s2 srt2 ) ->
+                    if c1 /= c2 || q1 /= q2 || f1 /= f2 || srt1 /= srt2 then
+                        submitQuery newModel ( newModel, cmd )
+
+                    else
+                        pair
+
+                _ ->
+                    pair
+    in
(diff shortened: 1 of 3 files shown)


garbas:
The hack is not completely removed. In `src/Main.elm`

Also removing the `#disabled` from the url now triggers the request to the backend when those links are clicked, which is not the case on master (you can compare it with search.nixos.org).

src/Main.elm
```diff
@@ -349,52 +322,37 @@ view model =
...
-                    model.query
-                        |> Maybe.map (\q -> [ Url.Builder.string "query" q ])
-                        |> Maybe.withDefault []
-                        |> List.append [ Url.Builder.string "channel" model.channel ]
-
-                Options model ->
-                    model.query
-                        |> Maybe.map (\q -> [ Url.Builder.string "query" q ])
-                        |> Maybe.withDefault []
-                        |> List.append [ Url.Builder.string "channel" model.channel ]
```
turboMaCk, line 367: there is actually bug with my previous PR. I was not aware that there is another place where we build up URL. I think the decoding/encoding of search query gets broken when pages got switched. Sorry I haven't caught this before - it should have occurred to me that there must a be some mechanism for preserving query prams when changing the routes.

src/Main.elm
```diff
@@ -49,7 +49,7 @@ type alias Flags =
 
 type alias Model =
     { navKey : Browser.Navigation.Key
-    , url : Url.Url
+    , route : Route.Route
```
turboMaCk, line 52: I know that keeping the route as well as page type seems like a duplicating the information. Anyway storing route is better (more type-safe) than storing url. I also think even NoRedInk stopped using Sum type style `Page` type because it's hard to work with and use product (record with multiple nested models). For now there is this slight duplication of information but swith to route helped me to catch one bug and made some parts of implementation simpler.

src/Main.elm
```diff
@@ -158,80 +158,74 @@ changeRouteTo :
...
     -> ( Model, Cmd Msg )
-changeRouteTo model url =
+changeRouteTo currentModel url =
     let
-        cleanUrl =
-            if url.fragment == Just "disabled" then
-                { url | fragment = Nothing }
+        attempteQuery (( newModel, cmd ) as pair) =
+            case ( currentModel.route, newModel.route ) of
+                ( Route.Packages c1 q1 _ f1 s1 srt1, Route.Packages c2 q2 _ f2 s2 srt2 ) ->
```
garbas, line 165: please use full names for variables

src/Main.elm
```diff
@@ -158,80 +158,74 @@ changeRouteTo :
...
+                    if c1 /= c2 || q1 /= q2 || f1 /= f2 || srt1 /= srt2 then
+                        submitQuery newModel ( newModel, cmd )
 
-            else
-                url
+                    else
+                        pair
+
+                ( Route.Options c1 q1 _ f1 s1 srt1, Route.Options c2 q2 _ f2 s2 srt2 ) ->
```
garbas, line 172: please use full names for variables

garbas (approved):
I just have some naming comments but other then that it looks and works perfectly.

PR #221, opened 2020-11-02
Closes #220 Latest stable is now the default channel

### LINKED ISSUE #220: Change default channel to latest-stable
When opening either [packages](https://search.nixos.org/packages) or [options](https://search.nixos.org/options), the default channel is "unstable". I think that making the latest stable release (e.g.: 20.09 as of right now) the default channel would be a more reasonable default for most of the users.

plfaucher:
* Declared a variable `latest_stable` pointing to the latest stable channel near channels definition. (Line 293)
* Used the variable on line 124 for `defaultChannel` when unspecified.

This is imho better than only changing the line `Maybe.withDefault "unstable"` to `Maybe.withDefault "20.09"`, because if done this way we would have to change version numbers at two places (Line 124 and the definitions around Line 280)

### DIFF
--- a/src/Search.elm
+++ b/src/Search.elm
@@ -121,7 +121,7 @@ init args model =
         defaultChannel =
             model
                 |> Maybe.map (\x -> x.channel)
-                |> Maybe.withDefault "unstable"
+                |> Maybe.withDefault latest_stable
 
         defaultFrom =
             model
@@ -287,6 +287,11 @@ channelDetails channel =
         Release_20_09 ->
             ChannelDetails "20.09" "20.09" "nixos/release-20.09" "nixos-20.09"
 
+-- Concentrates logic near releases definition.
+-- Otherwise you would have remember to modify @:124 defaultChannel each time.
+-- Bump when new channel is released.
+latest_stable : String
+latest_stable = "20.09"
 
 channelFromId : String -> Maybe Channel
 channelFromId channel_id =
--- a/src/Search.elm
+++ b/src/Search.elm
@@ -118,30 +118,30 @@ type Sort
 init : Route.SearchArgs -> Maybe (Model a) -> ( Model a, Cmd (Msg a) )
 init args model =
     let
-        defaultChannel =
+        channel =
             model
                 |> Maybe.map (\x -> x.channel)
-                |> Maybe.withDefault latest_stable
+                |> Maybe.withDefault defaultChannel
 
-        defaultFrom =
+        from =
             model
                 |> Maybe.map (\x -> x.from)
                 |> Maybe.withDefault 0
 
-        defaultSize =
+        size =
             model
                 |> Maybe.map (\x -> x.size)
                 |> Maybe.withDefault 30
     in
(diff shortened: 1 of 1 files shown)


src/Search.elm
```diff
@@ -287,6 +287,11 @@ channelDetails channel =
         Release_20_09 ->
             ChannelDetails "20.09" "20.09" "nixos/release-20.09" "nixos-20.09"
 
+-- Concentrates logic near releases definition.
+-- Otherwise you would have remember to modify @:124 defaultChannel each time.
+-- Bump when new channel is released.
+latest_stable : String
+latest_stable = "20.09"
```
turboMaCk, line 294: I think we should do this more programmatically. It would be more work but also require less maintenance.

garbas, line 294: lets rename this into `defaultChannel`, but it would be actually nice to change how we handle channels in a way that we wouldn't have to update this every time the release is made.

turboMaCk: I agree. Though I'm not against merging this PR as first step and then continuing with more future proofed solution in new PR.

garbas: We could fetch active channels via the metrics, like status.nixos.org already does it (https://github.com/NixOS/nixos-org-configurations/blob/master/delft/eris/status-page/status.js#L102) and then pass this list of string down (flake.nix -> src/index.js -> src/Search.elm)

plfaucher: `defaultChannel` is already defined on line 121 for the request model unfortunately. 

You are both 100% right that it should not be a manual update in the code. I view default channel issue and the automatic channel fetching as two separates enhancements, so if you agree I would:

1. Create a separate issue for the channel fetching. {Someone} can then work on it. I am very much a novice in Elm but I wouldn't mind trying :)
2. Rename `latest_stable` to `defaultChannelStable` or anything else to your liking that will not conflict with existing code.

garbas (approved):

src/Search.elm
garbas: I renamed the latest_stable to defaultChannel and merge the PR since it is a step forward.

If you would like to work on this I can offer help to mentor you. The task will require a bit more then just Elm. Here is how things should be done (in order):

- create `./update.sh` where we will fetch active channels ([like status.nixos.org](https://github.com/NixOS/nixos-org-configurations/blob/master/delft/eris/status-page/status.js#L102)) and store then in `./channels.json`
- add update step to cron workflow (`.github/workflows/cron.yml`). we also need to commit back the changes like we do [in nixos-homepage](https://github.com/NixOS/nixos-homepage/blob/master/.github/workflows/cron.yml#L48)
- in `default.nix` read `channels.json` and pass make it available in the dev/build environment via variable like we already do this for `ELASTICSEARCH_MAPPING_SCHEMA_VERSION`.
- in `src/index.js` we would use just passed variable as a initialization flag, like we already do for `ELASTICSEARCH_MAPPING_SCHEMA_VERSION`.
- we would extend Flags alias type in `src/Main.elm` and pass it down to `src/Search.elm`

PR #222, opened 2020-11-08
recurse into haskell.compiler for to find additional Haskell compilers

adamse:


garbas:
@adamse thank you. It looks good, just need to check if it works once I have some time.

garbas:
Thank you!

PR #223, opened 2020-11-16
Remove unnecessary jquery-ui reference

BenediktSeidl:
This file seems to be missing since 2020-09-22:
https://web.archive.org/web/20200922051906/https://nixos.org/js/jquery-ui.min.js

As there are no bug reports since then, I think it may be save to remove the reference?

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -6,12 +6,8 @@
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <title>NixOS Search - Loading...</title>
 
-
-  <script type="text/javascript" src="https://nixos.org/js/jquery.min.js"></script>
-
   <meta name="viewport" content="width=device-width, initial-scale=1.0" />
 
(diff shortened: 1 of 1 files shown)


turboMaCk:
I'm not sure if we even want to have jquery there. Even dynamic bootstrap stuff won't work witl elm's vdom I believe.

BenediktSeidl:
> I'm not sure if we even want to have jquery there. Even dynamic bootstrap stuff won't work witl elm's vdom I believe.

grepping over the project it seems like jquery is not in use so i removed it. the bootstrap script throws an error, so i was bold enough to remove this, too. everything seems to work fine.

it seems like font-awesome is also not in use, as i could not see any fonts loaded in the developer tools, so i removed this, too.

turboMaCk:
yes the twitter bootstrap (ui framework) comes with some JS which depends on jquery. Anyway I think this JS won't work with elm anyway so I think removing it is a good approach. Anyway this will need to be tested.

garbas:
I'm fixing the github actions in #227  and then i'll merge this.

PR #230, opened 2020-11-19
Refactor the loading and request management

### LINKED ISSUE #224: clicking the search button for a already loaded result shows inifite loading animation
### how to reproduce

* surf to https://search.nixos.org/packages?channel=20.09&from=0&size=30&sort=relevance&query=nixos
* click "Search" button
* see loading animation that never resolves to search result

### expected result

either
* fetching the search result again from api
* do nothing, but keep the previous search result

*tested the currently active version on https://search.nixos.org/ with chromium `80.0.3987.162` and firefox `68.8.0esr`*

turboMaCk:
resolves #224

TODO:

- [x] Unify the source of truth between loading and request sending
- [x] Update pagination for compatibility with new approach
- [x] Update sorting for compatibility with new approach

This changes the way we handle Loading:

- `result : RemoteData` is now source of truth for if we should or should not trigger the search
- `resut = Loading` is set only when right conditions are met (eg term is not empty)
- `Main.elm` queries data only when result state is `Loading`
- Unlike before, we now need to set `Loading` also for pagination. Because of this I made a small layout changes so that pagination is visible even in loading state for better UI.

This PR could be improved further - for instance we can avoid having `prev` button enabled during loading etc. However I didin't want to introduce way too many changes in single PR.

### DIFF
--- a/src/Main.elm
+++ b/src/Main.elm
@@ -108,50 +108,47 @@ updateWith toPage toMsg model ( subModel, subCmd ) =
     )
 
 
-submitQuery :
-    Model
-    -> ( Model, Cmd Msg )
-    -> ( Model, Cmd Msg )
-submitQuery old ( new, cmd ) =
+attemptQuery : ( Model, Cmd Msg ) -> ( Model, Cmd Msg )
+attemptQuery (( model, _ ) as pair) =
     let
-        triggerSearch _ newModel msg makeRequest =
-            if
-                (newModel.query /= Nothing)
-                    && (newModel.query /= Just "")
-                    && List.member newModel.channel Search.channels
-            then
-                ( new
-                , Cmd.batch
-                    [ cmd
-                    , makeRequest
-                        new.elasticsearch
-                        newModel.channel
-                        (Maybe.withDefault "" newModel.query)
-                        newModel.from
-                        newModel.size
-                        newModel.sort
-                        |> Cmd.map msg
-                    ]
-                )
+        -- We intentially throw away Cmd
+        -- because we don't want to perform any effects
+        -- in this cases where route itself doesn't change
+        noEffects =
+            Tuple.mapSecond (always Cmd.none)
+
+        submitQuery msg makeRequest searchModel =
+            Tuple.mapSecond
+                (\cmd ->
+                    Cmd.batch
+                        [ cmd
+                        , Cmd.map msg <| makeRequest
+                            model.elasticsearch
+                            searchModel.channel
+                            (Maybe.withDefault "" searchModel.query)
+                            searchModel.from
+                            searchModel.size
+                            searchModel.sort
+                        ]
+                ) pair
+    in
+    case model.page of
+        Packages searchModel ->
+            if Search.shouldLoad searchModel then
+                submitQuery PackagesMsg Page.Packages.makeRequest searchModel
 
             else
-                ( new, cmd )
-    in
-    case ( old.page, new.page ) of
-        ( Packages oldModel, Packages newModel ) ->
-            triggerSearch oldModel newModel PackagesMsg Page.Packages.makeRequest
+                noEffects pair
 
-        ( NotFound, Packages newModel ) ->
-            triggerSearch newModel newModel PackagesMsg Page.Packages.makeRequest
+        Options searchModel ->
+            if Search.shouldLoad searchModel then
+                submitQuery OptionsMsg Page.Options.makeRequest searchModel
 
-        ( Options oldModel, Options newModel ) ->
-            triggerSearch oldModel newModel OptionsMsg Page.Options.makeRequest
+            else
+                noEffects pair
 
-        ( NotFound, Options newModel ) ->
-            triggerSearch newModel newModel OptionsMsg Page.Options.makeRequest
-
-        ( _, _ ) ->
-            ( new, cmd )
+        _ ->
+            pair
 
 
 changeRouteTo :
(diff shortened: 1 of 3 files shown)


src/Main.elm
```diff
@@ -108,100 +108,56 @@ updateWith toPage toMsg model ( subModel, subCmd ) =
...
-                ( new, cmd )
+                pair
     in
-    case ( old.page, new.page ) of
-        ( Packages oldModel, Packages newModel ) ->
-            triggerSearch oldModel newModel PackagesMsg Page.Packages.makeRequest
+    case model.page of
+        Packages searchModel ->
+            if Search.shouldLoad searchModel then
+                submitQuery PackagesMsg Page.Packages.makeRequest searchModel
```
turboMaCk, line 140: single source of truth for when to perform loading.

src/Main.elm
```diff
@@ -263,7 +219,13 @@ update msg model =
...
                 Browser.External href ->
                     ( model
-                    , Browser.Navigation.load href
+                    , case href of
+                        -- ignore links with no `href` attribute
+                        "" ->
+                            Cmd.none
+
+                        _ ->
+                            Browser.Navigation.load href
```
turboMaCk, line 228: this fixes issues with links lacking `Attributes.href` reloading app.

src/Search.elm
```diff
@@ -148,10 +149,25 @@ init args model =
...
+    model.result == RemoteData.Loading
+
+
+ensureLoading : Model a -> Model a
+ensureLoading model =
+    if model.query /= Nothing && model.query /= Just "" && List.member model.channel channels then
+        { model | result = RemoteData.Loading }
+
+    else
+        model
```
turboMaCk, line 168: this is the only place where we explicitely define in which case we are able to load the data.

src/Search.elm
```diff
@@ -522,138 +543,128 @@ view { toRoute, categoryName } title model viewSuccess outMsg =
...
+                                remainder =
+                                    if remainderBy model.size total == 0 then
+                                        1
+
+                                    else
+                                        0
+                            in
+                            ChangePage <| ((total // model.size) - remainder) * model.size
+                    ]
+                    [ text "Last" ]
```
turboMaCk, line 665: most of the diff here is due to changes in indentation due to `Html.map`.

src/index.less
```diff
@@ -1,6 +1,7 @@
 body {
   position: relative;
   min-height: 100vh;
+  overflow-y: scroll;
```
turboMaCk, line 4: This is to prevent page jumping few pixels in width. Having results might force scrollbars while loading state does not (list is very short). Setting vertical scrollbar to be always visible means no changes in document width as these two state change.

src/index.less
```diff
@@ -89,6 +90,21 @@ header .navbar.navbar-static-top {
...
+  & > li > a {
+    cursor: pointer;
+    margin: 0 2px;
+  }
+}
+
+.loader-wrapper {
+  height: 200px;
+  overflow: hidden;
+}
```
turboMaCk, line 107: this is to prevent jumping due to changes in height as animation is running

src/index.less
```diff
@@ -89,6 +90,21 @@ header .navbar.navbar-static-top {
...
+.sort-form,
+.sort-form > .sort-group {
+  margin-bottom: 0;
+}
+.pager {
+  & > li > a {
+    cursor: pointer;
+    margin: 0 2px;
+  }
+}
```
turboMaCk, line 102: Sorry I could not resist my OCD alarm. This is just to clean up the layout just a tiny bit.

**before**
image


**now:**
image

garbas (changes requested):
Just did a quick test and I noticed that when you click on a result item a new request is sent to the backend. I think we messed up this once and I'm not sure I'm going to remember to check this always. We should make this somehow more reliable, but I have zero idea how.

src/Main.elm
```diff
@@ -108,100 +108,56 @@ updateWith toPage toMsg model ( subModel, subCmd ) =
     )
 
 
-submitQuery :
-    Model
-    -> ( Model, Cmd Msg )
-    -> ( Model, Cmd Msg )
-submitQuery old ( new, cmd ) =
+attemptQuery : ( Model, Cmd Msg ) -> ( Model, Cmd Msg )
+attemptQuery (( model, _ ) as pair) =
```
garbas, line 112: Didn't know you can use `as` in this way. Neat!

src/Main.elm
```diff
@@ -263,7 +219,13 @@ update msg model =
 
                 Browser.External href ->
                     ( model
-                    , Browser.Navigation.load href
+                    , case href of
+                        -- ignore links with no `href` attribute
+                        "" ->
```
garbas, line 224: Will this work properly also for links with "#someid"?

src/Search.elm
garbas: I think some condition is missing since we shouldn't load when we open details of an search result item.

garbas:
There is also another thing that it would be nice to do, while we are at it. When a result item is selected, we should scroll to it. I think this was working at some point or maybe I was just dreaming that I implemented it :) This is in a reference to this bug report #231

src/Search.elm
turboMaCk: in that case we simply don't call this function.

turboMaCk:
good catch.

The problem is:

- open detail triggers urlChange effect
- effects is handled in Main
- in main `init` is called which clears the results

These sum type page models are quite tricky to work with :cold_sweat: . We need to prevent init from being called. in situations like this.

garbas (approved):

Issue #105, opened 2015-12-01
html double escape in options.html


cleverca22:
the javascript in https://github.com/NixOS/nixos-homepage/blob/master/nixos/options.tt#L233 appears to escape < and > twice

an example can be seen on https://nixos.org/nixos/options.html by searching for services.znc.confOptions.passBlock

garbas:
https://search.nixos.org/options?show=services.znc.confOptions.passBlock&query=services.znc.confoptions.passblock&from=0&size=15&sort=relevance&channel=unstable#disabled

ncfavier:
After discussing with @Infinisil, this is a mistake on nixpkgs' side, nothing to do with nixos-search.

PR #236, opened 2020-11-26
Fix rendering of default and example values

### LINKED ISSUE #161: Newlines in code blocks are rendered as \n
The example for the NixOS option `nixpkgs.overlays` reads
```nix
let # fix the syntax highlighting by making this a valid Nix expression
      example = literalExample
        ''
          [
            (self: super: {
              openssh = super.openssh.override {
                hpnSupport = true;
                kerberos = self.libkrb5;
              };
            })
          ]
        '';
```
however this is rendered as

test

Link to the search: https://search.nixos.org/options?show=nixpkgs.overlays&query=overlays&from=0&size=15&sort=relevance&channel=unstable#disabled

### LINKED ISSUE #175: Code blocks may render as invalid Nix expression
The default value for the [`nix.nixPath`](https://github.com/NixOS/nixpkgs-channels/blob/c59ea8b8a0e7f927e7291c14ea6cd1bd3a16ff38/nixos/modules/services/misc/nix-daemon.nix#L413-L418) option is
```nix
let # fix the syntax highlighting by making this a valid Nix expression
        default =
          [
            "nixpkgs=/nix/var/nix/profiles/per-user/root/channels/nixos"
            "nixos-config=/etc/nixos/configuration.nix"
            "/nix/var/nix/profiles/per-user/root/channels"
          ];
```
However, this is rendered as

test

The commas between the elements are not valid Nix.

Link to the search: https://search.nixos.org/options?show=nix.nixPath&query=nixPath&from=0&size=30&sort=relevance&channel=unstable#disabled

### LINKED ISSUE #233: Structured option defaults and examples are rendered as JSON
```
10:13 <Reventlov> hey there
10:13 <Reventlov> why ther are commas in https://search.nixos.org/options?channel=20.09&show=services.unbound.interfaces&from=0&size=30&sort=relevance&query=services.unbound.interfaces ?
10:13 <Reventlov> i.e. Default value ["127.0.0.1", "::1"]
```
https://search.nixos.org/options?channel=20.09&show=services.unbound.interfaces&from=0&size=30&sort=relevance&query=services.unbound.interfaces

ncfavier:
This commit:

- Removes the `json.dumps` call around literal examples (this fixes #161).
- Converts non-literal examples and default values from JSON to valid Nix values by calling `nix-instantiate` (this fixes #175 and fixes #233). It's probably not very efficient to call `nix-instantiate` once per option, but I couldn't find a better way.
- Fixes some typoes.

Note that #105 is not related to this and should ~~probably~~ be closed, as the HTML escapes are already present in the NixOS module (https://github.com/NixOS/nixpkgs/blob/release-20.09/nixos/modules/services/networking/znc/options.nix#L173-L179).
EDIT: after discussing with the author, this was indeed a mistake.


garbas:
@ncfavier thank you! I'll look how costly running `nix-instantiate` is and if it is in the limits of acceptable I'll merge it otherwise we'll need to find a better option.

garbas:
I merged current master and bumped the `VERSION` which should allow us to test this. I'm running the import script manually now for the `20.09` channel.

ncfavier:
I think if we ① memoize the calls to `toNix` and ② avoid calling `nix-instantiate` for simple values (numbers, empty things, maybe strings), the time should be acceptable.

I'll commit soon.

ncfavier:
The calls to `nix-instantiate` seem to be taking around 1 minute total, now.

We could probably go further and implement the translation manually in Python (the only hard part is strings, I think). This would allow to print multi-line strings with the `''...''` syntax, as is done in the manual.

ncfavier:
This last commit addresses two other issues:

- some options, like `nixpkgs.pkgs`, use `literalExample` in their `defaultText` setting and are therefore rendered incorrectly (this is probably a mistake on the nixpkgs side, but it doesn't hurt to accept it here)
- default values set to `null` are currently reported as "Not given" (for example `users.motd`)

garbas:
I noticed that you used some python 3.9 feature. for now we are still using 3.8, you can enter the development environment with:

```
$ nix develop ./#packages.x86_64-linux.import_scripts
```

And you can run the import to your elasticsearch instance by running 

```
$ nix build -v ./#packages.x86_64-linux.import_scripts && ./result/bin/import-channel --es-url=$ELASTICSEARCH_URL --channel=20.09 -vvv -f
```

ncfavier:
Ah yes, sorry, I wasn't testing locally so far. Will do.

garbas:
@ncfavier no worries, if you have time jump to `#nixos-marketing` channel and we can debug it together.

PR #242, opened 2020-12-01
Fix package position links

ncfavier:
Splitting at the 44th character means leaving the `source/` part which is not part of GitHub nixpkgs URLs. All "Nix expression" links in the package search are currently broken because of this.

This fix strips the first three path components instead (`/nix/store/...-source`).


ncfavier:
Oops, I tried bumping the version in `VERSION` to trigger the import but now it 404s. I'm just gonna wait.

garbas:
@ncfavier no need to bump the version, this is only needed when we change elasticsearch schema.

ncfavier:
Noted.

garbas:
@ncfavier Sorry I wrote to fast before. The VERSION change lets us test - alongside existing search index.

ncfavier:
Ok, I assume it's something **you** have to do?

garbas:
yes, sadly, It would be nice to automate this somehow.

import-scripts/import_scripts/channel.py
```diff
@@ -462,7 +462,7 @@ def gen():
 
             position = data["meta"].get("position")
             if position and position.startswith("/nix/store"):
-                position = position[44:]
+                position = position.split("/", 4)[-1]
```
garbas, line 465: I think some time ago this was actually working. but then something changed in nixpkgs and `source/` was added. It think we need to be a little more defensive here. I would only remove `source/` and nothing else. We definetly dont want to rely on structure here.

```
position = position[44:]
if position.startswith('source/'):
    position = position[len('source/'):]
```

or something similar

ncfavier: But then things will break again if `source/` changes to `src/` or whatnot. I think we can reasonably rely on the fact that it will always look like `/nix/store/hash-name/actual_path_in_nixpkgs`, and therefore stripping the first three path components should be enough.

garbas (approved):

garbas:
I'll import the channels manually and then merge this.

PR #246, opened 2020-12-06
import-scripts: fix boolean rendering

ncfavier:
Apparently `isinstance(True, int)` is `True` in Python, and so booleans are rendered as Python values, with a capital letter. I changed the conditions to test against `type(value)` instead.


garbas (approved):
:rocket:

PR #248, opened 2020-12-06
Fix redirect from / to /packages breaking browser back button

bmwalters:
Previously, navigating to https://search.nixos.org/ would use history.pushState
to redirect to /packages and add a new entry to the browser history.
This meant that if you pressed the back button, you would navigate back to /
which would of course then redirect to /packages again.

Now history.replaceState is used instead of pushState, so the / history entry
is replaced entirely by a /packages history entry and the back button works.

### DIFF
--- a/src/Main.elm
+++ b/src/Main.elm
@@ -202,7 +202,7 @@ changeRouteTo currentModel url =
                 Route.Home ->
                     -- Always redirect to /packages until we have something to show
                     -- on the home page
-                    ( model, Browser.Navigation.pushUrl model.navKey "/packages" )
+                    ( model, Browser.Navigation.replaceUrl model.navKey "/packages" )
 
                 Route.Packages searchArgs ->
                     let


bmwalters:
For some reason the Netlify deployment is failing even after I add NETLIFY_AUTH_TOKEN and NETLIFY_SITE_ID to my repo's secrets?

Could be because of this line? https://github.com/NixOS/nixos-search/blob/27ff900fc31a5a06bac3ce1e5b1a1b6506afea97/.github/workflows/main.yml#L56

bmwalters:
https://5fcd57a4ded7dd08c3f57095--gifted-villani-93c941.netlify.app/

garbas:
@bmwalters thank you for fixing this. Using `replaceUrl` makes more sense here.

The preview links are only generated when created from the branch on this repo. If you are planning to contribute more to `nixos-search` I can give you the permission to push branches, but if not I'd like to keep committers list as small as possible.

garbas (approved):

PR #249, opened 2020-12-07
Add retry to parts of the import script which were failing from time to time

### LINKED ISSUE #243: Options search doesn't work for 20.09
At the moment, options search for 20.09 doesn't find anything. I tried different keywords that definitely exist, like "services" or "hardware" - always get the same. 

_search action returns a valid object with no errors but with zero results:

```json
{
  "took":0,
  "timed_out":false,
  "_shards":{
    "total":1,
    "successful":1,
    "skipped":0,
    "failed":0
  },
  "hits":{
    "total":{
      "value":0,
      "relation":"eq"
    },
    "max_score":null,
    "hits":[]
  }
}
```

Example URL: https://search.nixos.org/options?channel=20.09&from=0&size=30&sort=relevance&query=services

garbas:
fixes #243

### DIFF
--- a/import-scripts/import_scripts/channel.py
+++ b/import-scripts/import_scripts/channel.py
@@ -1,6 +1,8 @@
+import backoff  # type: ignore
 import boto3  # type: ignore
 import botocore  # type: ignore
 import botocore.client  # type: ignore
+import botocore.errorfactory  # type: ignore
 import click
 import click_log  # type: ignore
 import dictdiffer  # type: ignore
@@ -14,6 +16,7 @@ import os.path
 import pypandoc  # type: ignore
 import re
 import requests
+import requests.exceptions
 import shlex
 import subprocess
 import sys
@@ -262,6 +265,7 @@ def parse_query(text):
     return tokens
 
 
+@backoff.on_exception(backoff.expo, botocore.errorfactory.NoSuchKey)
 def get_last_evaluation(prefix):
     logger.debug(f"Retrieving last evaluation for {prefix} prefix.")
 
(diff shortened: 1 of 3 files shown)


garbas:
I think this should make the import script more robust.

import-scripts/import_scripts/channel.py
```diff
@@ -521,10 +528,11 @@ def get_options(evaluation):
     options = get_options_raw(evaluation)
 
     @functools.lru_cache(maxsize=None)
+    @backoff.on_exception(backoff.expo, subprocess.CalledProcessError)
     def jsonToNix(value):
         result = subprocess.run(
             shlex.split(
-                "nix-instantiate --eval --strict -E 'builtins.fromJSON (builtins.readFile /dev/stdin)'"
+                "nix-instantiate --eval --strict -E 'builtins.fromJSON (builtins.areadFile /dev/stdin)'"
```
ncfavier, line 535: Is this a typo?

ncfavier (approved):
Looks good, although it would be nice to know exactly why the command is failing. I'll try and reproduce the issue locally with a standard error log.

import-scripts/import_scripts/channel.py
garbas: oh right, good catch. this was to test if it is failing correctly.

garbas:
> Looks good, although it would be nice to know exactly why the command is failing. I'll try and reproduce the issue locally with a standard error log.

The problem is intermittent, I couldn't reproduce it even once locally.

ncfavier:
> The problem is intermittent, I couldn't reproduce it even once locally.

OK

ryantm:
I currently see no search results for 20.09 or unstable options.

garbas:
@ryantm I hoped simply adding a retry would solve this, but it looks like we need to either revert #239 or properly implement it. And it wouldn't hurt to make the import script more robust.

PR #254, opened 2020-12-17
Improve nix.py

ncfavier:
Adds escaping of `\` and `"` inside double-quoted strings (see `system.name`), and fixes indentation for double-single-quoted strings and `literalExample`s (see `services.snapper.configs`).


garbas:
@ncfavier can you add some tests for this use cases? Also if you can point to me which options are being fixed by this?

ncfavier:
I did: `system.name` and `services.snapper.configs`. Where would you like me to add tests?

garbas (approved):
@ncfavier no worries. I've merged the master and added the tests.

I will run the import manually in few minutes.

ncfavier:
Looks good to me!

PR #259, opened 2021-01-01
Make search result elements keyboard focusable

### LINKED ISSUE #258: Search result elements should be keyboard focusable
Currently, none of the search results can be navigated or selected with keyboard input alone, because they lack the HTML tags that tell the browser how to select them.

Adding the HTML attributes `tabindex` and `role` (button) to the heading of each search result would allow these headings to be expanded and contracted without the use of a mouse, which will enable more users to make better use of the site.

jakeisnt:
Closes #258 
- Replaced `<tr/>` element with `<a/>`
- Styled `<a/>` like `<tr/>`

This may not be a comprehensive fix, but it works well with Vimium.
If there are any style conventions I should follow please let me know!

### DIFF
--- a/src/Page/Options.elm
+++ b/src/Page/Options.elm
@@ -33,6 +33,7 @@ import Html.Attributes
         ( class
         , colspan
         , href
+        , style
         , target
         )
 import Html.Events
@@ -174,8 +175,11 @@ viewResultItem channel show item =
         -- DEBUG:         ]
         -- DEBUG:     ]
         |> List.append
-            (tr
-                [ onClick (SearchMsg (Search.ShowDetails item.source.name))
+            (a
+                [ style "display" "table-row"
+                , style "color" "unset"
+                , style "text-decoration" "none"
+                , onClick (SearchMsg (Search.ShowDetails item.source.name))
                 , Search.elementId item.source.name
                 ]
                 [ td [] [ text item.source.name ]
--- a/src/Page/Packages.elm
+++ b/src/Page/Packages.elm
@@ -34,6 +34,7 @@ import Html.Attributes
         ( class
         , colspan
         , href
+        , style
         , target
         )
 import Html.Events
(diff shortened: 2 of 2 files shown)


jakeisnt:
I realized I'd only added this fix to `Packages`; just added it to `Options` as well.

garbas (changes requested):
@jakeisnt thank you for taking the time to fix this. 

The HTML you are producing is not correct, since you can only have `td` inside `tr`. I think we would need to move away from `table` anyway and use [CSS Grid](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout).

Also please put the styles into `src/index.less`.

turboMaCk:
I did not tried it but I think the simplest solution would be to just add `tabindex` attribute to `tr` element (which has onClickHandler) to make it focusable by tab. If this works we avoid any changes in HTML and CSS to get this functionality.

jakeisnt:
Did you try the changes out? `display: table-row` causes the `<a/>` tag to display like a `<tr/>` tag, so this is perfectly fine. With this change the website appears identical to how it did before. 

I'm not sure what you mean by 'not correct' but this is definitely a valid solution to the issue. 

I agree that inline styles aren't the best practice - I cam move those.

@samueldr mentioned in the issue I opened that an html tag should be used rather than `tabindex` as the browser already defines behavior for interacting with HTML tags and tabindex can be finicky.

samueldr:
My suggestion wrapped the first cell's content:

 - https://github.com/NixOS/nixos-homepage/blob/b60ed4af1f3b94b9415312249a8725fbff6ab2aa/packages-explorer/src/gui/result.js#L33-L40

(Here it's a button, but an `<a>`nchor can also be used if we have a real href to point it to)

The main drawback with wrapping only the first element with a link is that the "whole row" won't be selected by keyboard navigation, but that's not really an issue.

turboMaCk:
@samueldr actually regarding:

> The main drawback with wrapping only the first element with a link is that the "whole row" won't be selected by keyboard navigation, but that's not really an issue.

What if we would do exactly that and just remove the "only". We can keep both button as well as event handler on `<tr>`. That way tab:

- for mouse users we would keep convenience of having whole row clickable
- tab selection would simply work via buttons
- to meet the accessibility standard we would have button as well (an element that is expected to have action on click).

samueldr:
> What if we would do exactly that and just remove the "only". We can keep both button as well as event handler on <tr>. That way tab

To remove the "only", three scenarios:

 - wrap the whole `<tr>` → broken HTML.
 - stop using a `<table>` → need to re-implement styles (possible with upcoming redesign).
 - add as many `<button>` as columns → need to tab through the whole row before going to the next.

But, uh, now that I look at your solution, what you implemented is what I had in mind, and I see the miscommunication. I didn't intend to make *only* the first cell *clickable*, I was only thinking about adding *one* focusable element per row.

garbas:
This has been solved in #260. @jakeisnt thank you for bringing this issue up.

PR #260, opened 2021-01-05
Add semanticaly clickable `button` element to result list

### LINKED ISSUE #189: Impossible to Expand Search Results With Keyboard
For accessibility reasons, it's important that websites can be navigated and used fully without a mouse. There are many technologies that help people do this: frame navigation, caret navigation, screenreaders, and vimium's link hints mode. They all, however, require that the data they need be present on the page in some way. The results table of nixos-search presents as a plain table with option names and click handlers. Even if I use caret navigation to put my text cursor directly on the option name and hit enter, the click handler doesn't fire and I have no way to get at the information within. I'm not super familiar with the web, so I'm not sure what the right answer is here, but links would be awesome because those are super easy to find and activate.

### LINKED ISSUE #258: Search result elements should be keyboard focusable

turboMaCk:
- improves usability with keyboard #258
- improves accessbility

See #259 for other relevant discussion

keyboard-and-click

* * *

 * Closes #258 
 * Closes #189

### DIFF
--- a/src/Page/Options.elm
+++ b/src/Page/Options.elm
@@ -154,6 +154,9 @@ viewResultItem channel show item =
 
             else
                 []
+
+        open =
+            SearchMsg (Search.ShowDetails item.source.name)
     in
     []
         -- DEBUG: |> List.append
@@ -175,10 +178,22 @@ viewResultItem channel show item =
         -- DEBUG:     ]
         |> List.append
             (tr
-                [ onClick (SearchMsg (Search.ShowDetails item.source.name))
+                [ onClick open
                 , Search.elementId item.source.name
                 ]
-                [ td [] [ text item.source.name ]
+                [ td []
+                    [ Html.button
+                        [ class "search-result-button"
+                        , Html.Events.custom "click" <|
+                            Json.Decode.succeed
+                                { message = open
+                                , stopPropagation = True
+                                , preventDefault = True
+                                }
+                        ]
+                        [ text item.source.name
+                        ]
+                    ]
                 ]
                 :: packageDetails
             )
--- a/src/index.less
+++ b/src/index.less
@@ -94,6 +94,10 @@ header .navbar.navbar-static-top {
       padding: 0;
       border: 0;
       background: transparent;
+      display: inline;
+      font: inherit;
+      color: inherit;
+      text-align: inherit;
   }
 }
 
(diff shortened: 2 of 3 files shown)


src/Page/Options.elm
```diff
@@ -175,10 +178,22 @@ viewResultItem channel show item =
...
                 ]
-                [ td [] [ text item.source.name ]
+                [ td []
+                    [ Html.button
+                        [ class "search-result-button"
+                        , Html.Events.custom "click" <|
+                            Json.Decode.succeed
+                                { message = open
+                                , stopPropagation = True
+                                , preventDefault = True
```
turboMaCk, line 191: - preventDefault is to avoid page refreshes
- stopPropagation is to avoid bubbling to parent element which would caused second message which would immediately close the item again

samueldr:
(Edited the description to auto-close relevant issues.)

src/index.less
```diff
@@ -88,6 +88,13 @@ header .navbar.navbar-static-top {
   tbody > td > dl > dd {
     margin-bottom: 1em;
   }
+
+  .search-result-button {
+      margin: 0;
+      padding: 0;
+      border: 0;
+      background: transparent;
+  }
```
samueldr, line 97: https://github.com/NixOS/nixos-homepage/blob/b60ed4af1f3b94b9415312249a8725fbff6ab2aa/packages-explorer/src/styles/app.less#L80-L85

Not sure `text-aiign` is needed here, but `font` and `color` would be. `display` I'm not sure either, might end up being a no-op with bootstrap.

```
	display: inline;
	font: inherit;
	color: inherit;
	text-align: inherit;
```

Why `inherit` font and color? Because form elements in HTML do not use the body's styles by default, but the user-agent/system's "chrome" styles. Let's imagine the user uses a dark theme with white button text, in the past (if it is still not an issue) in this situation the button text color would be white.

samueldr (changes requested):

samueldr:
Good work. Details to fixup in the styles, but otherwise appropriate approach.

garbas (approved):

Issue #201, opened 2020-10-01
Use nixpkgs as name of the channel for example how to install it


garbas:
and mention nixos channel in case you are on NixOS

More from the twitter discussion https://twitter.com/zekjur/status/1311617711068377089

stapelberg:
Specific replies that deserve attention (also because Twitter makes them hard to browse):

https://twitter.com/ioPhileas/status/1311581193062830080

> True, I'd be happy if it would say: "channel nixos not found, available channels: nixpkgs"

https://twitter.com/domenkozar/status/1311567418033045505

> We're running a campaign to improve such error messages https://opencollective.com/nix-errors-enhancement
> 
> In short, it's due to channels abstraction anti-pattern that is going away in Nix 3.

turboMaCk:
resolved by https://github.com/NixOS/nixos-search/pull/261

Issue #95, opened 2020-06-12
Search input, channels and other search options


garbas:
Currently a user needs to:
1. Select if they want to search packages or options
2. Select the channel (or just leave the default)
3. Input their query

But going forward I would like to simplify even more how a user searches and add much more search options. Doing this blindly and just adding check boxes will lead to a poor design. And this is the reason for this issue, to collect all ideas, search options and looking into future how to design best user experience.

Things to consider in out discussion:

- "Show unfree packages" checkbox option
- "Show broken packages" checkbox option
- "Licences" multi select option
- "Maintainers" multi select option
- "Package sets" multi select option (pythonPackages, haskellPackages, rPackages, ...)
- Do different types of sorting make sense? Alphabetically? or is relevance search the only one we actually care
- To not "scare" the user we can only show only few options next to search input and once search happens we can show the rest of the options hidden behind the "Advance search options..." link.
- Can we make only one search which yield results for for different types: Packages, Options, Issues, Documentation, ...  We could again do something similar as google does it, where you can see tabs of different type of results *after* you made the search, eg: All, Images, Maps, News, ...

turboMaCk:
I think it might be a good idea talk about this before the redesign. I think the idea is great. The only point I'm a bit afraid is selection of maintainer which might be quite tricky ux-wise provided how many there are and how much this can grow in the future. I would probably suggest de-scoping it from first iteration on this. I'm not sure about priority of including issues and documentation in first iteration as well since these also can be extracted as separate epic imo.

garbas:
@turboMaCk if you have time come to `#nixos-marketing` channel and we can have scope this correctly.

turboMaCk:
resolved by https://github.com/NixOS/nixos-search/pull/261

PR #261, opened 2021-01-07
Adds faceted search

### LINKED ISSUE #163: faceted search for packages
Initially there only a simple search bar with 3 channels is visible.

Add and "advanced" link which will open the rest of criteria that you can filter on. This advance bar should be also visible once search is performed.

Criteria we should add:
- package set (python27Packages, vimPackages, nodePackages, emacsPackages, ...)
- license
- maintainer
- platform

garbas:
This PR add faceted (filtered) search for packages.

At the same time it changes the layout of the results, from table to ordered list. While the design is not final, the structure of the result items should give hint how this is going to be styled with https://github.com/nixos/nixos-common-styles.

A lot of inspiration was taken from github's search -> https://github.com/search

Fixes #163  #261 #95 #201


src/index.less
```diff
@@ -59,6 +59,24 @@ header .navbar.navbar-static-top {
...
 }
+
+.search-faceted {
+  margin-top: 6.5em;
+
+  & > ul {
+    padding: 0;
+  }
+  
+  ul.nav-stacked > li > a {
```
turboMaCk, line 70: I think using class as oppose to structure might scale better but it can be improved later,

turboMaCk (commented):
@garbas let me know if I can (once I can) help with anything.

garbas:
> @garbas let me know if I can (once I can) help with anything.

I need to fix the query a bit but then I would really welcome the help how to make faceted search easier to use. I need to do a bit of a research in this area.

samueldr:
Can we have a description to the PR explaining what is the intent behind the change? So we can validate it works as intended? :)

garbas:
> Can we have a description to the PR explaining what is the intent behind the change? So we can validate it works as intended? :)

Oh sorry, I thought I opened a draft. Anyway add the description now.

garbas:
@samueldr @turboMaCk This is now ready for review.

samueldr:
Changing the **Sort** criteria affects the URL, but does not change it in the combo box, and *does not seem to affect the results*. Though, **refreshing** with the URL affected seems to work. Looks like something isn't causing the refresh of the results to be sent.

 * [ ] (check when fixed)

* * *

The clickable area for a result is... *odd*...

image

In green is the clickable area.

Suggestions:

 - On hover add a background to the whole `li.option` to make it more obvious it is clickable
 - Change from `margin-bottom: 2em` to `padding-top: 2em`. (You may need to adjust the listing top and bottom margin.)

 * [ ] (check when fixed)

* * *

Platforms listing isn't filtered to useful platforms anymore.

This makes for useless verbosity:

image

The current search filters platforms for those that are... pragmatically useful? At least those with binary cache coverage.

 * [ ] (check when fixed)

* * *

Package metadata is hard to understand. I **guess** that it is *attribute name*, *version*, for the first elements, then the *license*, *homepage* and *source* links are oddly not link-like.

Two main issues, first is there is no labeling on what the *thing* is, then there is no way to know that a thing is a link without hovering.

Sadly, I don't really know how to better solve it for being useful *at a glance*. The table-like layout used previously really is the pragmatic approach, it is hard to misunderstand.

image

 * [ ] (check when fixed)

samueldr:
The clickable area for a result is... *odd*...

image

In green is the clickable area.

Suggestions:

 - On hover add a background to the whole `li.option` to make it more obvious it is clickable
 - Change from `margin-bottom: 2em` to `padding-top: 2em`. (You may need to adjust the listing top and bottom margin.)

 * [ ] (check when fixed)

samueldr:
Platforms listing isn't filtered to useful platforms anymore.

This makes for useless verbosity:

image

The current search filters platforms for those that are... pragmatically useful? At least those with binary cache coverage.

 * [ ] (check when fixed)

samueldr:
Package metadata is hard to understand. I **guess** that it is *attribute name*, *version*, for the first elements, then the *license*, *homepage* and *source* links are oddly not link-like.

Two main issues, first is there is no labeling on what the *thing* is, then there is no way to know that a thing is a link without hovering.

Sadly, I don't really know how to better solve it for being useful *at a glance*. The table-like layout used previously really is the pragmatic approach, it is hard to misunderstand.

image

 * [ ] (check when fixed)

samueldr:
I assume all narrow-viewport issues are to be looked at with the redesign. It's hard to fix all of this in a narrow layout.

Though, if not prohibitively expensive, it would be nice to have the side bar (with the facets) somehow be a togglable thing when narrow.

(No actual *need* to fix before merging, but something to keep in mind.)

samueldr:
Having an item opened at the bottom of the page, then changing facets scrolls to the opened item.

I guess that comes from the handling of navigation to a *deep* URL. I don't know how easy this is going to be to fix. To reproduce:

 * https://5ffc4a201e18091ff840ec31--nixos-search.netlify.app/packages?channel=unstable&show=rPackages.pcaBootPlot&from=0&size=50&buckets=%7B%22package_attr_set%22%3A%5B%22rPackages%22%2C%22haskellPackages%22%5D%2C%22package_license_set%22%3A%5B%5D%2C%22package_maintainers_set%22%3A%5B%5D%2C%22package_platforms%22%3A%5B%5D%7D&sort=relevance&query=boot
 * Change facets on the left.

 * [ ] (check when fixed)

samueldr:
It is basically impossible to select information in an opened item. This is because the selection click, on release, hides the item. Not useful to select those installation instructions!

I guess the solution is somewhat linked to the problematic activation area. It might be necessary to do like the current implementation, and only close when the always present "header" is clicked. To help users, a "useless" "× close" button could be added too.

 * [ ] (check when fixed)

src/Page/Options.elm
```diff
@@ -111,126 +111,64 @@ update navKey msg model =
 view : Model -> Html Msg
 view model =
     Search.view { toRoute = Route.Options, categoryName = "options" }
-        "Search NixOS options"
+        [ text "Search more than "
+        , strong [] [ text "10 000 options" ]
```
samueldr, line 115: Any strategies to help ensure this number doesn't drift too far?

src/Page/Packages.elm
```diff
@@ -148,259 +190,357 @@ update navKey msg model =
 view : Model -> Html Msg
 view model =
     Search.view { toRoute = Route.Packages, categoryName = "packages" }
-        "Search NixOS packages"
+        [ text "Search more than "
+        , strong [] [ text "80 000 packages" ]
```
samueldr, line 194: Any strategies to help ensure this number doesn't drift too far?

samueldr:
Don't let this shower of papercuts discourage you. In reality these are good steps towards improvement.

garbas:
@samueldr Thank you a ton! I really appreciate the deep review. I'll try to address reported issues today or tomorrow. Thank you again.

turboMaCk:
First I went through @samueldr's comments so I don't repeat what he said. Just quickly.

- [ ] Changing the Sort criteria doesn't work - reproduced **[high]**
- [ ] The clickable area for a result is... odd... (non intuitive) - reproduced **[low]**
- [ ] Platforms listing in a result isn't filtered to useful platforms anymore. - wasn't able to reproduce **[unknown]**
- [ ] Package metadata is hard to understand. - disagree, I think this is fine. **[none]**
- [ ] I assume all narrow-viewport issues are to be looked at with the nixos-styles redesign - agree can waut **[low]**
- [ ] Having an item opened at the bottom of the page, then changing facets scrolls to the opened item. - reproduced but will need some additional definition - worth doing in isolation (own PR) imo - **[low]**
- [ ] It is basically impossible to select information in an opened item. This is because the selection click, on release, hides the item. Not useful to select those installation instructions! - reproduced - **[high]**

## Additional Issues from my own testing:

### No packages found usability

When no package is found the filters are not shown. This means that if I select combination that will result in 0 results I endup in "bricked like" state. Clicking `Search` btn again fixes this but it's really not obvious how to get out. Solution to this will require some thinking but I feel we should probably keep previous filters states during loading.

image

#### To reproduce:

1. search `elm` term
1. filter `package set: elmPackages`
1. filter `maintainers: Lynn Dong`

src/Page/Options.elm
turboMaCk: I think there are essentially two ways.

- dynamic - fetch this information over network
- static - inline this information during build

I think @garbas will know what is the right call more than me. However I agree that hard-coding this is won't be scalable. But I'm ok resolving this in separate issue/PR.

src/Page/Options.elm
```diff
@@ -111,126 +111,64 @@ update navKey msg model =
...
+        viewBuckets
         SearchMsg
 
 
+viewBuckets :
+    Maybe String
+    -> Search.SearchResult ResultItemSource ResultAggregations
+    -> List (Html Msg)
+viewBuckets _ _ =
+    []
```
turboMaCk, line 128: I think putting `(\_ _ -> [])` directly on the line 119 would make it more obvious that we don't expect anything here in this case and that this is intentional.

src/Page/Options.elm
```diff
@@ -111,126 +111,64 @@ update navKey msg model =
...
-                    (viewResultItem channel show)
-                    result.hits.hits
-                )
-            ]
-        ]
+viewSuccess channel showNixOSDetails show hits =
+    ul []
+        (List.concatMap
+            (viewResultItem channel showNixOSDetails show)
+            hits
```
turboMaCk, line 141: I think using `List.map` together with changing API of `viewResultItem` to return `Html msg` would be better here. Concatination of lists here is completely unnecessary

`concatMap` (as well as `concat` and `append`) has a lot of overhead - it's in fact orders of magnitude slower than folding of `(++)` over the list due to implementations details in `elm/core` - we still wait for improvement in core sadly. It's maybe not a big deal as we don't expect to have tens of thousands of items here but considering view function is called many times over the lifetime of application (after every Msg) and that this code will run on many machines I think we can save some wats over time.

src/Page/Options.elm
```diff
@@ -271,21 +201,54 @@ viewResultItemDetails channel item =
...
-        , dd [] [ withEmpty (wrapped asPreCode) item.source.example ]
-        , dt [] [ text "Declared in" ]
-        , dd [] [ withEmpty asGithubLink item.source.source ]
+    [ li
+        [ class "option"
+        , onClick open
+        , Search.elementId item.source.name
         ]
+        ([]
+            |> List.append showDetails
```
turboMaCk, line 239: This is completely unnecessary as it is equivalent to:


```suggestion
        (showDetails
```

Following the monoid identity law of list

> There exists an element e in S such that for every element a in S, the equations e • a = a and a • e = a hold.

in this case `e = []` and `• = List.append`

src/Page/Options.elm
```diff
@@ -271,21 +201,54 @@ viewResultItemDetails channel item =
...
+                    , div [] [ withEmpty asPre item.source.type_ ]
+                    , div [] [ text "Example" ]
+                    , div [] [ withEmpty (wrapped asPreCode) item.source.example ]
+                    , div [] [ text "Declared in" ]
+                    , div [] [ withEmpty asGithubLink item.source.source ]
+                    ]
+                ]
+
+            else
+                []
```
turboMaCk, line 228: Not sure if it's worth it but I would consider rather returning `Maybe (Html msg)` here and using `List.filterMap identity` to get rid off nothing. I think there is even function for this in `list-extra`.

src/Page/Packages.elm
```diff
@@ -148,259 +190,357 @@ update navKey msg model =
...
+        |> viewBucket
+            "Platforms"
+            (result.aggregations.package_platforms.buckets |> List.sortBy .doc_count |> List.reverse)
+            (createBucketsMsg .platforms (\s v -> { s | platforms = v }))
+            selectedBucket.platforms
+        |> viewBucket
+            "Maintainers"
+            (result.aggregations.package_maintainers_set.buckets |> List.sortBy .doc_count |> List.reverse)
+            (createBucketsMsg .maintainers (\s v -> { s | maintainers = v }))
+            selectedBucket.maintainers
```
turboMaCk, line 243: I would consider simplifying this writter pattern into `List.filterMap` together with chage in api of `viewBucket` function to return maybe.

src/Page/Packages.elm
```diff
@@ -148,259 +190,357 @@ update navKey msg model =
...
-                    [ th [] [ text "Attribute name" ]
-                    , th [] [ text "Name" ]
-                    , th [] [ text "Version" ]
-                    , th [] [ text "Description" ]
-                    ]
+    -> List Search.AggregationsBucketItem
+    -> (String -> Msg)
+    -> List String
+    -> List (Html Msg)
+    -> List (Html Msg)
```
turboMaCk, line 252: here we could return the `Maybe (Html Msg)` as described in comment above.

src/Page/Packages.elm
```diff
@@ -148,259 +190,357 @@ update navKey msg model =
...
-                :: packageDetails
-            )
+    -> List (Search.ResultItem ResultItemSource)
+    -> Html Msg
+viewSuccess channel showNixOSDetails show hits =
+    ul []
+        (List.concatMap
+            (viewResultItem channel showNixOSDetails show)
+            hits
+        )
```
turboMaCk, line 300: another instance of the `concatMap`

src/Page/Packages.elm
```diff
@@ -148,259 +190,357 @@ update navKey msg model =
...
+            |> List.append longerPackageDetails
+            |> List.append
+                [ Html.button
+                    [ class "search-result-button" ]
+                    [ text item.source.attr_name ]
+                , div [] [ text <| Maybe.withDefault "" item.source.description ]
+                , shortPackageDetails
+                ]
+        )
+    ]
```
turboMaCk, line 543: another instance of unnecessary singleton list

src/Search.elm
```diff
@@ -435,253 +537,400 @@ view :
...
+                                            ]
+                                        , onClick <| outMsg (ChannelChange channel.id)
+                                        ]
+                                        [ text channel.title ]
+                                )
+                    )
+                    channels
+                )
+            ]
+        ]
```
turboMaCk, line 732: I can't put suggestion in but this can be simplified to using `cons` on a list instead.

`(div [] .....) :: (List.fiterMap)`

src/Search.elm
```diff
@@ -435,253 +537,400 @@ view :
...
+
+                 else
+                    model.from + model.size
+                )
+
+        total =
+            String.fromInt result.hits.total.value
+    in
+    [ div []
+        [ viewSortSelection toRoute model
```
turboMaCk, line 777: another singleton list

src/Search.elm
```diff
@@ -435,253 +537,400 @@ view :
...
+--                            ]
+--                            [ text <| toSortTitle sort ]
+--                    )
+--                    sortBy
+--                )
+--            ]
+--        ]
+--    ]
+
+
```
turboMaCk, line 872: perhaps this can be removed?

src/Search.elm
```diff
@@ -788,74 +1039,62 @@ makeRequestBody query from sizeRaw sort type_ sortField fields =
...
+                                                ]
+                                            )
+                                          ]
+                                        ]
+                                  )
+                                ]
+                          )
+                        ]
+                  )
+                ]
```
turboMaCk, line 1077: Also can use cons instead of append of singleton list.

src/Utils.elm
```diff
@@ -0,0 +1,13 @@
...
+
+
+toggleList : 
+    List a
+    -> a 
+    -> List a
+toggleList list item =
+    if List.member item list
+    then List.filter (\x -> x /= item) list
+    else List.append list [item]
```
turboMaCk, line 13: Assuming we don't care about order, this would be much better:

```suggestion
    else item :: list
```

Otherwise it would be far better to use `Set` for data we need to perform these operations with. Set itself is contained to `comparable` types but I wrote [any-set](https://package.elm-lang.org/packages/turboMaCk/any-set/latest/) to solve this.

turboMaCk (commented):
This is huge huge improvement. Great work getting this off the ground.

I've commented on some minor issues in code. I think neither of them is blocker. As I see it there are some issues which needs to be addressed before we can put this out (see my previous comment). Improvements can come later.

turboMaCk:
sorry. I pushed to this branch by mistake - now reset back to where this should be.

turboMaCk:
some improvements around clicks within result list. https://github.com/NixOS/nixos-search/pull/262

turboMaCk:
RE reported issue. I'm able to reproduce issue with long list of platforms. `nginx` package is one such example

image

src/Page/Options.elm
garbas: Ideally this would be done as a counter that counts from 0 to a number which is fetched from the backend meanwhile. But for now this was the easier hack.

src/Page/Options.elm
garbas: I've added this as a placeholder. Also you wouldn't know what that function should return if we would make it anonymous.

src/Page/Packages.elm
garbas: Ideally this would be done as a counter that counts from 0 to a number which is fetched from the backend meanwhile. But for now this was the easier hack.

src/Search.elm
garbas: :+1:

garbas:
I fixed the issues, at least I think so.

@turboMaCk I haven't yet address any of the elm code style issues. We can do this after this PR is merged.

turboMaCk:
@garbas Don't worry these are mostly just notes. I don't want to block merge of this longer than absolutely necessary

turboMaCk:
Just a tiny cleanup after myself - now only the button (with name) is clickable https://github.com/NixOS/nixos-search/pull/263

turboMaCk:
There is a regression in options search

image


The styles meant for "show more" button affects the display of code examples here. CSS selector as they're written now are quite hard to get right. It's not really clear what selector will match what. I think we should use more classes and less styling based on HTML structure.

turboMaCk:
fix for the options displaying incorrectly... We will eventually need to refactor those css though. This was harder to fix than it suppose to be. We can do it as a part of redesign or later. https://github.com/NixOS/nixos-search/pull/264

turboMaCk:
Also. Once these 2 PRs are merged it's approve :+1: from my side.

turboMaCk (approved):
:heavy_check_mark: I've tested this again one more time just to be sure.

LGTM! :tada: 

The only thing I'm a bit worried about is that this makes it again a bit harder to use with keyboard which is something some folks were unhappy previously. But I don't feel that should be blocker. Keyboard navigation key can be improved later.

Overall this looks and feels like a huge step forward. Great work @garbas

garbas:
Before I would merge this I would like to make the mobile version of it a bit more mobile friendly. Kinda following the same ideas as on https://github.com/search

And regarding the keyboard navigation it works as good as github's search, at least as far as I can tell. But yes there is more links you need to tab over.

src/Page/Options.elm
garbas: There is a Count API that we can use from elasticsearch https://www.elastic.co/guide/en/elasticsearch/reference/current/search-count.html

turboMaCk:
Another smaller CSS issue which we should preferably solve 
image

garbas:
- [ ] Another thing which we should add is also "No package set", "No platform", "No licences" and "No mainatainers"  option (this would partially solve https://github.com/NixOS/rfcs/pull/81)

- [ ] Another things is we should handle platforms at the import and limit to only platforms we are interested in.

samueldr:
Platform support tiers:

 - https://github.com/NixOS/rfcs/pull/46

samueldr:
Off-by-one in "showing results" subtitle:

https://6004568634bcdafcc4b348ec--nixos-search.netlify.app/packages?channel=20.09&from=0&size=50&sort=relevance&query=fire

 - Showing results 0-50 of 164 packages.
 - Showing results 50-100 of 164 packages.
 - Showing results 100-150 of 164 packages.
 - Showing results 150-164 of 164 packages.

Should be:

 - Showing results 1-50 of 164 packages.
 - Showing results 51-100 of 164 packages.
 - Showing results 101-150 of 164 packages.
 - Showing results 151-164 of 164 packages.

Think about result 50, is it shown on the first and second page? no :)

hugolgst (commented):
Here are my few comments:
- The search icon looks weird for me: 

- I find it confusing that the "Show more detail" text is on the separator:

- Could be great to underline the items where you can click, otherwise, it won't be obvious:
image
- After clicking a title, the focus outline stays:
image

Otherwise, it looks great!

turboMaCk:
@hugolgst which browser is it?

This is how it looks for me in FF

nixos-search

hugolgst:
I'm using google-chrome in i3-wm (don't know if this has a role)

garbas:
> Another smaller CSS issue which we should preferably solve

This is now fixed

garbas:
> Here are my few comments:
> 
>     * The search icon looks weird for me:
>       
> 

I'm not sure how to do this better without importing additional font. I would leave this for now since I plan to work on redesign next. And there we will make sure this kind of things don't happen.

>     * I find it confusing that the "Show more detail" text is on the separator:
>       
> 

I also find it confusing. But I don't have better idea - at this point - how to solve it. As for the previous item I would table this until the redesign.

>     * Could be great to underline the items where you can click, otherwise, it won't be obvious:
>       image
> 

This is somehow connected with the previous item since it is about displaying all the fields of a package. When redesigning (implementing the same design as nixos.org) I will look how to improve this.

>     * After clicking a title, the focus outline stays:
>       image
> 

I'm also not able to reproduce this outline. I wonder if this is something specific to you. Can you debug it and see which style change would fix it?

garbas:
@hugolgst 
> > ```
> > * After clicking a title, the focus outline stays:
> >   image
> > ```
> 
> I'm also not able to reproduce this outline. I wonder if this is something specific to you. Can you debug it and see which style change would fix it?

I just noticed you saif you were using chrome. I can see this being a problem there. let me fix it.

hugolgst:
Thanks for the answers @garbas.
For the last point I added to fix it. 
```css
outline: none;
```

turboMaCk:
@hugolgst 

> Thanks for the answers @garbas.
> For the last point I added this CSS property to fix it.
> ```
> outline: none;
> ```

This change is undesirable because outline is needed for keyboard navigation. See https://github.com/NixOS/nixos-search/issues/258

garbas:
Regarding the mobile stuff I will implement them also with the redesign. At the current state it would just take to long to implement this properly. The sooner I merge this the sooner I will start working on redesign.

turboMaCk:
> I'm also not able to reproduce this outline. I wonder if this is something specific to you. Can you debug it and see which style change would fix it?

This should be reproducible in any chromium based browser which all have default rule for this.

image

While I agree it doesn't look nice I think we would went into rabbit hole if we would try to fix all these things within the scope of the PR.

I would suggest:

> - The search icon looks weird for me

It took me a while to figure out what this is about because from screenshot it was not apparent where is this `Q` thing placed. Anyway I got it.
image

I think this is yet another regression caused by https://github.com/NixOS/nixos-search/commit/4d505f34022154a1f193f012c11eb08d07c90271 (the font awesome removal). Adding font awesome will make `content: "\2315";` work in chrome.

> - I find it confusing that the "Show more detail" text is on the separator

I personally don't find it confusing but if some does it should be addressed. But I would take it as a consideration for redesign rather than try to design it within this PR

> - After clicking a title, the focus outline stays

I don't think we should try to solve these small things in scope of this. We need outline for keyboard navigation. I would suggest opening separate issue to improve outline styling just so we don't forget to make sure this is improved in scope of redesign.

turboMaCk:
I was looking into icon/font-face. It turned out that in FireFox this is using `FreeSherif` font face. It's GNU licensed font.

adding `font-family: FreeSherif` to the before element fixes the problem provided this font is available on host system. Full fix though would require us to create webfont (which I already did) and load so we can display this correctly on all systems. However our current build system (webpack) doesn't support that. There are in fact some traces of configuration for static files but other parts of missing. `yarn build` doesn't include any assets into `dist`. This means that fixing this would require changes in build configuration which makes this a bit of a scope creep.

PR #262, opened 2021-01-16
faceted search trap click

turboMaCk:
This is the quickfix for:

- Click actions propagation
- tiny improvement to visibility of clickable area.

## Small improvement to display of Options details

image

## Make additional thinks in detail clickable and selectable

search-selectable

## Change Cursor and underline text on hover

search-cursor

Especially the hover actions will need some additional improvements but I think this might be good enough to unblock the feature.

### DIFF
--- a/src/Page/Options.elm
+++ b/src/Page/Options.elm
@@ -136,7 +136,7 @@ viewSuccess :
     -> Html Msg
 viewSuccess channel showNixOSDetails show hits =
     ul []
-        (List.concatMap
+        (List.map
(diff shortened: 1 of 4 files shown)


garbas (approved):

turboMaCk:
Just a tiny update to the hover action https://github.com/NixOS/nixos-search/pull/263 <- sort of rollbacks few of changes I made since now only title is clickable - I think we should not confuse users by underlining headline unless something clickable is hovered

PR #273, opened 2021-01-25
Remove the defunct "burger menu" and always expand the navigation

andir:
I commonly run into a situation where the burger menu button is being
showed to me but it broke some time ago and doesn't expand the menu
anymore.

Since the snipped was probably mostly 1:1 translated from the old
nixos.org page (where we had more links than just the three we have now)
we can probably just remove it now. It is a net win as the feature was
broken and at least now you can use the website on mobile again.

### DIFF
--- a/src/Main.elm
+++ b/src/Main.elm
@@ -305,20 +305,10 @@ view model =
                 [ div [ class "navbar navbar-static-top" ]
                     [ div [ class "navbar-inner" ]
                         [ div [ class "container" ]
-                            [ button
-                                [ type_ "button"
-                                , class "btn btn-navbar"
-                                , attribute "data-toggle" "collapse"
-                                , attribute "data-target" ".nav-collapse"
-                                ]
-                                [ span [ class "icon-bar" ] []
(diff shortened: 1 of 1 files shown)


turboMaCk:
I've fixed the hamburger in https://github.com/NixOS/nixos-search/commit/1c1a3ca7d19595bc2a09737fca403b82c0d5998a

But it seems to be broken again. This is probably due to bad rebase of https://github.com/NixOS/nixos-search/commit/5bb94c9c92e55375ffb288a3ae33e3f4773d974b

garbas:
@andir The burger menu is visible for now this was fixed by @turboMaCk in #267. This PR just removes the working burger menu.

To be honest I have no opinion on this because this is going to be changed in the coming PRs:
- design is going to be adopted from nixos.org and we will use burger menu (if there are going to be multiple options)
- packages and options are going to be actually merged and there is going to be no top menu actually which makes this debate kinda obsolete.

Since I'm without opinion I'm merging this since I really appreciate you taking the time to actually submit a PR. Thank you, thank you, thank you!

Issue #272, opened 2021-01-25
improve usabilty of nixos option search


tklitschi:
The current [NixOS Search](https://search.nixos.org/options) you took a step in the wrong direction. 
The search the used to have a year ago was a really nice experience, now you have to scroll to see every Option making it difficult to search in a wide variety and can´t even copy the option #271. 
This really makes searching for NixOS options a pain.

turboMaCk:
hi @tklitschi 

Can you please clarify what you mean by this point?

> now you have to scroll to see every Option making it difficult to search in a wide variety

turboMaCk:
@tklitschi Thanks for updating the title to be Options specific. I agree that options is at the moment sadly neglected. I belive this is because nixos-search falls under the wings of marketing team efforts which are currently primarily focused on beginner experience and team expects the primary beginner use-case to be nix package manager on other Linux distributions / MacOS. that's being said I'm not a member of marketing team so further clarification has to come from @garbas 

If you have a sec you can have a look at https://github.com/NixOS/nixos-search/pull/275 or test this url https://600eaa161f0dc71a959aa5fb--nixos-search.netlify.app/packages - It includes my proposed improvements to copy/paste UX especially in options.

If you can expand on what you would like to see or don't like to see in Options search specifically please feel free to expand on that. I myself care about Options search experience and would like to look into this more in next several days in terms of opening PRs with minor improvements as well as coming up with issues for long term improvements (eg I would like to have option to filter by namespace).

tklitschi:
I guess the package search makes sense the way it is implemented now but since package and option search are completely different use cases it doesn't make any practical sense to use the same UI besides marketing. 
I really appreciate your effort in #275 and I guess it a step in right direction.

The big problem is, if I want to get for example into the nginx module and want to get an overview off all options the experience looks like this:
![Peek 2021-01-25 13-17](https://user-images.githubusercontent.com/51724453/105705688-d3a73c00-5f10-11eb-8a59-68d6de17709e.gif)

You get a lot of Information, some you don't need. What made the 'old' nixos-search nice to use, was that you just get the name and by that an idea what that option does and if you want to get detailed information, you could click to expand. 
![old-option](https://user-images.githubusercontent.com/91113/87112822-1ce1b800-c26d-11ea-9541-5dddea0c3e1d.png)
It was nice and compact and you could read everything directly.

piegamesde:
Closely related, the UX of the expand mechanism can be improved in many points:

[...]

PR #275, opened 2021-01-25
Improve copy/paste UX in Options and Packages

### LINKED ISSUE #271: Cannot select option names for easy copying
Before the recent update, the name of the option was one of the things shown in the details of the option. This was useful for selecting and copying, e.g. to share option names via IRC or for pasting into configs. Now the redundant copy is missing, and the option "link" can't be selected.

turboMaCk:
This improves the copy/paste experience especially for Options.

- Change title from button to a to make it selectable
- Resurrect `Name` meta for Options
- Make sure this won't break keyboard selection

# Item titles are now selectable (both packages and options):
selectable-title

*This was broken due to support for keyboard navigation (wrapping to button) https://github.com/NixOS/nixos-search/pull/260. Fixed in a way it's still focusable by keyboard as well as selectable**

# Resurrection o `Name` meta for Options

name-meta

Resolves #271 #272 #276

### DIFF
--- a/src/Page/Options.elm
+++ b/src/Page/Options.elm
@@ -213,7 +213,9 @@ viewResultItem channel _ show item =
         showDetails =
             if Just item.source.name == show then
                 div [ Html.Attributes.map SearchMsg Search.trapClick ]
-                    [ div [] [ text "Default value" ]
+                    [ div [] [ text "Name" ]
+                    , div [] [ wrapped asPreCode item.source.name ]
+                    , div [] [ text "Default value" ]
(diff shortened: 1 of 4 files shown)


lheckemann:
Thanks, much better! If the "title" is selectable, I don't think we need the meta field any more. A great bonus, if you could, would be adding a copy button :)

garbas:
I hijack this PR and also removed the description from the listing and showing it inside the details view.

garbas (approved):

PR #288, opened 2021-02-07
Improving search query

### LINKED ISSUE #218: Use version and attr_set as a second sort key
Example:
 - When searching for `firefox` we will match multiple version of firefox but show the latest version of firefox first.
 - When searching for `python requests` first result will be the requests library from latest python version

### LINKED ISSUE #226: 'pytorchWithCuda' missing, when 'pytorchWithoutCuda' exists
So confusingly `pytorchWithoutCuda` exists as a searchable package whereas `pytorchWithCuda` does not. The defintion of both of them is just 

```` 
pytorchWithCuda = self.pytorch.override { cudaSupport = true; };

pytorchWithoutCuda = self.pytorch.override { cudaSupport = false; };
````

so I am not sure why one is searchable and not the other.

### LINKED ISSUE #237: tensorflow_2 missing from search
Maybe it's related to https://github.com/NixOS/nixos-search/issues/226.

`tensorflow` is indexed in the search, but `tensorflow_1` or `tensorflow_2` aren't. The definitions in `python-packages.nix` look like:

```
    tensorflow = self.tensorflow_1;
    tensorflow_1 = self.tensorflow-build_1;
    tensorflow_2 = self.tensorflow-build_2;
```

### LINKED ISSUE #250: First query after reindexing takes ~30s
We need to warm up internal elasticsearch cache once everything has been imported. This can also serve as a quick test that everything works.

### LINKED ISSUE #279: Option search misses many results
Specifically, more general searches (shorter input) seem to return fewer results (at least sometimes) than more specific searches (longer input that is a superstring of the shorter input from before).

For a concrete example, take a search for [`ftp`](https://search.nixos.org/options?channel=unstable&from=0&size=500&sort=alpha_asc&query=ftp):

  * 12 hits
  * One of those is for options in `services.atftpd`
  * Three are for `services.vsftpd`

But if we search for [`services.atftpd`](https://search.nixos.org/options?channel=unstable&from=0&size=500&sort=alpha_asc&query=services.atftpd) we get three options, rather than the one we got before, and a search for [`services.vsftpd`](https://search.nixos.org/options?channel=unstable&from=0&size=500&sort=alpha_asc&query=services.vsftpd) returns 28 -- more than the entire previous search for `ftp` did, even though the results should be a strict subset of that search!

This appears to have happened along with the recent redesign of the search page that adds a lot more vertical padding to the search results, but that might be a coincidence.

garbas:
* Reduce the numbers of bucket results on each ChangeBucket event
* Warmup after reindexing, fixes #250
* Using multiple sort keys, fixes #218
* Index also unfree packages, fixes #226, fixes #237
* Add wilcard query, fixes #279


turboMaCk:
It seems that options search returns 400 on the deployment of this branch. https://601fcd5a8bacca822a330365--nixos-search.netlify.app/options?channel=20.09&from=0&size=50&sort=relevance&query=samba

garbas:
@turboMaCk Sorry I didn't see your comment, it only showed up after I merged. I'll check and come up with the fix ASAP

PR #289, opened 2021-02-07
Filter on buckets only when there is some bucket selected

garbas:
otherwise the query is malformed.

This fixes the bug found by @turboMaCk  in #288.


turboMaCk:
This seems to fix the problem indeed but for some reason it also cases Firefox to crash on my machine (happened both times I tried). It's most likely some sort of bug in FireFox or amdgpu and it didn't happen when I open search.nixos.org which means that I was maybe just unlucky. In case someone reports some similar mysterious issues let me know. I can try to reproduce it.

Otherwise thanks for the fix :+1:

PR #305, opened 2021-04-25
Fix `Q` icon font in before pseudo element

turboMaCk:
resolves 304

### DIFF
--- a/src/index.less
+++ b/src/index.less
@@ -107,6 +107,7 @@ header .navbar.navbar-static-top {
 
     &:before {
       content: "\2315";
+      font-family: FreeSerif, sans-serif;
       display: inline-block;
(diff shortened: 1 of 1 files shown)


samueldr:
I don't really think this is an appropriate solution.

This ends up (ab)using an implementation detail of the FreeSerif font, which may or may not be available on the end-user's machine.

It's using the *Telephone Recorder* character (??????) here.

 - https://www.compart.com/en/unicode/U+2315

Which many fonts implement with the line going through the circle (whatever it represents).

turboMaCk:
yes I think the the proper solution would be to include web fonts in here. I mentioned in past in some other issue but I think this would need some changes in build script and deployment.

PR #320, opened 2021-06-16
Rewrite import script in Rust

ysndr:


ysndr:
Status:

Packages:
- [x] import of flat flakes (with missing values)
- [x] parsing nixpkgs
- [ ]  representation of all keyword sets (attr_set, maintainers)
- [ ] attribute query

Options:
- [x] flake modules
- [ ] nixpkgs modules

Frontend:
Tbd

ysndr:
Options examples print verbatim currently

flake-info/src/bin/elastic-test.rs
```diff
@@ -0,0 +1,24 @@
...
+async fn check_index() -> Result<(),Box<dyn std::error::Error> > {
+
+    let client = elasticsearch::Elasticsearch::new(elasticsearch::http::transport::Transport::single_node("https://kog80y7vrg:myuaqqcb4r@nixos-search-5886075189.us-east-1.bonsaisearch.net:44")?);
+
+    let result = client
+            .index(IndexParts::Index("latest-21-21.05"))
+            .send()
+            .await?;
+            Ok(())
+}
```
garbas, line 24: I don't think this file is needed anymore

flake.nix
```diff
@@ -21,11 +21,22 @@
       packages = system:
         {
           import_scripts = mkPackage ./import-scripts system;
+          flake_info = mkPackage ./flake-info system;
           frontend = mkPackage ./. system;
         };
+
+      devShell = system:
```
garbas, line 28: :1st_place_medal:

src/Main.elm
```diff
@@ -362,6 +383,7 @@ viewNavigation route =
             (viewNavigationItem route)
             [ ( toRoute Route.Packages, "Packages" )
             , ( toRoute Route.Options, "Options" )
+            , ( toRoute Route.Flakes, "Flakes (Experimental)" )
```
garbas, line 386:
```suggestion
            --, ( toRoute Route.Flakes, "Flakes (Experimental)" )
```

Lets comment this for now

garbas (changes requested):

.github/workflows/cron-flakes.yml
```diff
@@ -58,11 +58,7 @@ jobs:
 
     - name: Import ${{ matrix.group }} group
       run: |
-        ./result/bin/flake-info --push --json group ./flakes/${{ matrix.group }}.json ${{ matrix.group }} 1>${{ matrix.group }}.json
-
-        # todo cache ES json?
-
-
+        ./result/bin/flake-info --push --elastic-schema-version=$(cat ./VERSION) group ./flakes/${{ matrix.group }}.json
```
ysndr, line 61: What is wrong about reading the version from the Environment?

garbas: You can not cat from file before checkout job :)

tomberek:
flake-info builds, runs, and functions as expected. Did a quick look through the added Rust; seems reasonable, no major red flags. (and i learned a few things! thank you).

How well isolated is this PR from the legacy portions of nixos-search? If there is high-confidence this will not disrupt the legacy portions, I recommend allowing @ysndr to merge/deploy/iterate on this. The current Nix roadmap suggests an upcomming release of 2.4 in the "next few weeks" and it would be nice to have this "live" by then.

ysndr:
The issue with Nixpkgs is that some modules pull insecure packages which causes the the nix script to fail
I can "fix" this by running the nix script with `NIXPKGS_ALLOW_INSECURE=1` 
`

ysndr:
@tomberek I'm quite confident this works equivalently to the current python script (except not filtering platforms, so all available systems are listed)

garbas (changes requested):
@ysndr we already chatted but this is really great work. I'll just summarize what still needs to be done before we merge this:

- [x] Filter platforms in packages search (do this in the frontend/elm). For now we should only show platforms that are supported and later we should figure out how to show other platforms.
- [x] Names of elastic index need to follow the following scheme -> `nixpkgs-<version>-<channel>-<git-rev>`, example `nixpkgs-21.05-123abc...`
- [ ] Latest index alias is not updated and should also be of new scheme `latest-nixpkgs-<version>-<channel>`
- [ ] Remove the `pull_requests:` from `.github/workflows/cron-nixpkgs.yml`
- [ ] Update exists strategy for `cron-nixpkgs.yml` to ignore.
- [ ] Fix Warmup job to use new index name

ysndr:
- update stRgety ignore means adding to an existing index, is this the right interpretation? I feel like aborting would be more suitable in this case wouldn't it?
- the nixpkgs tar all does not contain any git commit information.
  I can derive a version hash from the revs of the flakes, using nixpkgs would require either a different way to produce a hash or pulling it as a flake where the version could actually differ
- elastic search is giving me weird aggregations, i.e.missing platforms which do show in the individual items. Maybe an old index, will investigate further after aliasing is fixed.

ysndr:
`nix-env -f '<nixpkgs>' -I "https://github.com/NixOS/nixpkgs/archive/<ref>.tar.gz" -qa --json` 

produces different results on darwin than it does on linux
I assume it is because soem derivations fail to evaluate (`linux_rpi*`) packages are unavailable if evaluation happens on darwin!

garbas:
@ysndr yeah this is a known issue. I hope this will be possible in the future with flakes. For now we just need to get this rust version of the script on pair with the python one.

ysndr:
Well i wish i knew that earlier, took me quite a while and nerves to figure out
[unstable](https://6116657808e9ec1da01e8fca--nixos-search.netlify.app/packages?channel=unstable&from=0&size=50&sort=relevance&query=requests) and [21.05](https://6116657808e9ec1da01e8fca--nixos-search.netlify.app/packages?channel=21.05&from=0&size=50&sort=relevance&query=requests) are now showing the correct number of packages though

garbas (approved):

PR #324, opened 2021-06-25
Flake support/frontend

ysndr:


src/Search.elm
```diff
@@ -326,8 +354,8 @@ update toRoute navKey msg model =
 
         QueryResponse result ->
             -- let
-                -- _ =
-                --     Debug.log "got query result" result
+            -- _ =
+            --     Debug.log "got query result" result
             -- in
```
turboMaCk:
```suggestion
```

ysndr: I am aware of these and will remove all the debug lines soon before this gets merged

turboMaCk: yeah sorry. Frankly I mistaken the repo for something else.

tomberek:
Blocked by: https://github.com/NixOS/nixos-search/pull/320 ?

samueldr:
### Sidebar needs a minimum-width

<details>
<summary>screenshots</summary>

image

compared to:

image

</details>

In the package search this is probably a non-issue simply because the sidebar has more content, so the computed width ends up being better by default.

Note that the package search does not show a sidebar when there are no results, which means the behaviour *has* to differ between a null result set for flakes and non-flakes results (until the UX is totally reviewed/redone).

samueldr:
### Flake package search described as "options"

image

samueldr:
### Searching for flakes *themselves*

A user might want to find the flake itself, and not a package or option from said flake. E.g. I assume [`flake-utils`](https://github.com/numtide/flake-utils) couldn't show up here as it provides (AFAIUI) no package nor option.

Not 100% sure how to handle this. This should be *separate* from indexing utility functions. Finding a flake by name (or other meta information???) is, I would guess, a task in itself.

samueldr:
### Searching for utility functions (task)

How are utility functions handled generally in flakes? Taking [`flake-utils`](https://github.com/numtide/flake-utils) for example. How can this be indexed in a useful manner for end-users?

This is probably not relevant for this PR specifically and should be broken out into a future task to think and implement I figure.

ysndr:
Yes that part is planned (mid term) as i get all the info already (it's attached to all packages from a flake btw)Although we would need to gather some more attributes other than packagesApp attributes are already possible

ysndr:
Well the `lib` attribute is not in the specification so technically it's not handled at all

samueldr:
### Package/option toggle in narrow view (bug)

<details>
<summary>Screenshots</summary>

image

image

</details>

It's better for the package view, but for the options view it's really broken. This probably goes hand-in-hand with the previous issue reported, but just in case it doesn't.

Really an answer, which would help for the flake options listing view too, is to not add that "scope selector" into the sidebar? But I guess that goes with the UX review that needs to be done overall.

ysndr:
@samueldr @garbas **bug** reports should be fixed now

I will create issues for flakes (themselves) and additional attributes to be added later (utility/lib/app). Although I think this should wait until we have some unified interface.

samueldr:
### Flake package / options toggle shouldn't have a close icon (×) (bug)

image

This is because it is a binary choice between either. The close button being present on the option not currently selected has no use, and the close button on the selected option "does not work", it does not allow removing the filter.

samueldr:
Additionally, *Package/option toggle in narrow view* hasn't been fixed; 

<details>
<summary>Screenshot</summary>

image

</details>

The filter column still is extremely narrow, hiding all text.

PR #328, opened 2021-07-13
import-scripts: switch to poetry_core

jtojnar:
Full poetry is no longer needed: https://github.com/python-poetry/poetry-core


ysndr:
The import script is rewritten in rust with flake support in #320

PR #333, opened 2021-08-01
Style links in package details consistently

### LINKED ISSUE #331: Package Homepage, Source, Copyright Link Styling Unclear
There's no visual distinction in the `<ul>` of information about a package between name (not a link) and homepage (a link), as well as the copyright and source links. I'm new to NixOS, and spent about a week or so assuming I needed to separately duckduckgo a package to find its documentation.

I'd recommend making it the same blue as maintainers and platforms, and/or underlining the links. I'd also note that link underlining (on more than just hover, but as a default style) is always a bit clearer and more accessible.

waldyrious:
Making the links the same color as the non-linked text makes it hard to figure out which of the items are links and which are plain text.

This change lets the default link color apply to the links in package details.

Fixes #331.

| Before |
| ------ |
| Screenshot from 2021-08-01 10-43-12 |

| After |
| ----- |
| Screenshot from 2021-08-01 10-43-32 |

/cc @Fidgetcetera

### DIFF
--- a/src/index.less
+++ b/src/index.less
@@ -359,10 +359,6 @@ header .navbar.navbar-static-top {
               & > li {
                 display: inline-block;
                 margin-right: 1em;
-
-                & > a {
-                  color: #666;
-                }
               }
               & > li:last-child {
                 margin-right: 0;


turboMaCk:
I would be in favor of this change. What are your thoughts @garbas ?

ysndr:
Were just on the way to restructure the frontend in a bigger fashion in #324 to accommodate searchable flakes.Also there's the mid term goal to work on the style to match the UC of nixos.org

PR #343, opened 2021-08-23
Don't list unfree packages

andir:
There was a discussion around this when search.nixos.org was redesigned and the conclusion was that nixpkgs is a collection of free and open source software. We shouldn't list the unfree stuff.


flake-info/src/commands/packages-config.nix
```diff
@@ -3,7 +3,7 @@
   allowAliases = false;
 
   # Also list unfree packages
```
ncfavier, line 5: Don't forget to update the comment

ncfavier:
This actually didn't do anything, because of

https://github.com/NixOS/nixos-search/blob/a8275b15a190dcffc48788a31de196be930511d0/flake-info/src/commands/nixpkgs_info.rs#L78-L80

It seems like `allowUnfree` doesn't control what packages `nix-env` will list (it will always try to list everything), but rather whether it will fail to evaluate unfree packages. This should probably be implemented as an explicit filter somewhere else.

ncfavier:
Although now I think I'd actually oppose this change: this feels a lot like hiding unfree packages under the rug and seems disingenuous. If "nixpkgs is a collection of free and open source software" then it shouldn't include unfree software; otherwise, nixos-search shouldn't lie about what's included.

ysndr:
The environment variable does indeed affect the evaluation only,
I'm wondering why we have it here twice now as the unfree thing should have been/is set [here](https://github.com/NixOS/nixos-search/blob/a8275b15a190dcffc48788a31de196be930511d0/flake-info/src/commands/packages-config.nix#L3) too.

But I agree on having an explicit filter. This would also allow users to optionally include unfree packages in their search. I'm not sure if this was completely ruled out in the mentioned irc chat before but if it was I would be interested in the reason too ..

ncfavier:
> The environment variable does indeed affect the evaluation only, I'm wondering why we have it here twice now as the unfree thing should have been/is set [here](https://github.com/NixOS/nixos-search/blob/a8275b15a190dcffc48788a31de196be930511d0/flake-info/src/commands/packages-config.nix#L3) too.

It's not the same thing twice: `packages-config.nix` *disables* unfree whereas the environment variable *allows* it (and since the two are ORed together, the environment wins).

> But I agree on having an explicit filter. This would also allow users to optionally include unfree packages in their search. I'm not sure if this was completely ruled out in the mentioned irc chat before but if it was I would be interested in the reason too ..

Yes, I think that's the best compromise: don't list unfree packages *by default* but allow the user to tick a box.

ncfavier:
Uhm, actually the code I linked above is for options, not packages. The reason `nix-env` doesn't fail to evaluate unfree packages currently is that it doesn't need to evaluate them to get their metadata. This will change in the future when we add `--out-path` to support showing a package's outputs, since that requires evaluation, so we will then need to add the environment variables to the `nix-env` call too.

ysndr:
> This will change in the future when we add --out-path to support showing a package's outputs

Evaluating these will run into interruptions due to broken packages I assume, wouldn't it?

ysndr:
Yes thats what i was talking about
initially i patched that for a different approach with the options
seems like there is work towards that.


Although, given the new `nix` command I'd love to see rather the output for
```
nix search nixpkgs ".*" --json
// or 
nix flake show [--legacy] --json 
```
to be a bit more thorough instead


But maybe that's wishful thinking for now

Issue #303, opened 2021-04-15
DocBook rendering is broken


ncfavier:
The rendering of DocBook option descriptions to HTML implemented in #86 using Pandoc has multiple problems. The ones I could find are all conveniently illustrated in the description for [environment.profileRelativeSessionVariables](https://search.nixos.org/options?channel=unstable&show=environment.profileRelativeSessionVariables&from=0&size=30&sort=relevance&query=environment.profileRelativeSessionVariables):

1. ~cross-reference (`xref` elements) titles are not resolved properly and are replaced with `???`. This is because the cross-referenced element isn't present in the snippet we feed to Pandoc.~
2. ~cross-reference links just point to `#${id}`, which has no effect on Elasticsearch. They should probably point to the appropriate anchor in [options.html](https://nixos.org/manual/nixos/stable/options.html), or to a search.nixos.org URL such as the one I crafted above, or we should somehow teach Elasticsearch to deal with those anchors.~
3. ~man page references (`citerefentry` elements) aren't rendered properly. Pandoc [doesn't seem to support](https://github.com/jgm/pandoc/blob/2d60524de43d59ffb1763a33a15cc2ecce613ecf/src/Text/Pandoc/Readers/DocBook.hs#L99) that element at all at the moment.~
4. descriptions aren't broken into paragraphs at `\n\n`, like they are in the manual. This is because this is part of the post-processing that happens in [`optionsDocBook`](https://github.com/NixOS/nixpkgs/blob/081b9423d4ae18836c7564b634627b214049f80e/nixos/lib/make-options-doc/default.nix#L191-L192) but not in `optionsJSON`.

I'm not sure how we should fix this. The ways I can think of are:

1. Parse the `options.html` generated by `manualHTML` directly. This avoids a lot of code duplication, descriptions are rendered exactly as they are in the manual. On the downside, it requires parsing HTML and making assumptions about the generated format (default values, examples etc. aren't properly separated from the description text; CSS classes used in the manual might clash with those used by Elasticsearch).
[...]

PR #349, opened 2021-08-26
Add pandoc support back to rust script

ysndr:
Addressing #329 

Pandoc support for descriptions was removed in #320 and now comes back to the rewrite


jtojnar:
The [nixpkgs PR](https://github.com/NixOS/nixpkgs/pull/130047) has been merged.

ysndr:
Great, thanks for the notice

ysndr:
Internal links show as `???` still
Styling runs through pandoc with corresponding performance hit

jtojnar:
Yes, we did not fix the question marks yet. In the NixOS manual, we convert them [back to `xref` tags](https://github.com/NixOS/nixpkgs/blob/master/maintainers/scripts/doc/replace-xrefs-by-empty-links.py) and let DocBook resolve the chapter titles.

I think it should be possible to use manual’s olink database to get this information:

```patch
--- a/nixos/doc/manual/default.nix
+++ b/nixos/doc/manual/default.nix
@@ -159,7 +159,7 @@ let
     '';
 
 in rec {
-  inherit generatedSources;
+  inherit generatedSources olinkDB;
 
   inherit (optionsDoc) optionsJSON optionsXML optionsDocBook;
 
--- a/nixos/release.nix
+++ b/nixos/release.nix
@@ -132,6 +132,7 @@ in rec {
   manualEpub = (buildFromConfig ({ ... }: { }) (config: config.system.build.manual.manualEpub));
   manpages = buildFromConfig ({ ... }: { }) (config: config.system.build.manual.manpages);
   manualGeneratedSources = buildFromConfig ({ ... }: { }) (config: config.system.build.manual.generatedSources);
+  manualOlinkDB = buildFromConfig ({ ... }: { }) (config: config.system.build.manual.olinkDB);
   options = (buildFromConfig ({ ... }: { }) (config: config.system.build.manual.optionsJSON)).x86_64-linux;
 
 
```

ysndr:
seems like pandoc hasnt implemented olinks support yet..

> [ ] olink - A link that addresses its target indirectly, through an entity

https://github.com/jgm/pandoc/blob/37e30560ad546055d17f56df468cd35976b3218a/src/Text/Pandoc/Readers/DocBook.hs#L287

ysndr:
Is this at least the same state as  prior to the rewrite?

ncfavier:
Yes, see #303.

ysndr:
Ok, I'd say we should restore that state first.
The further issue can be handled in a separate PR

jtojnar:
`olink`s are not used in the manual anyway. But the olink database created by docbook-xsl can be used to resolve `xref`s as well. We would need to parse the olinkDB and replace the link texts in a lua script similar to https://github.com/NixOS/nixpkgs/blob/master/maintainers/scripts/doc/replace-xrefs-by-empty-links.py.

ysndr:
Ah I see
I think we can do that rather easily indeed, although I tried not to involve any xml parsing yet

.github/workflows/cron-nixpkgs.yml
```diff
@@ -4,6 +4,7 @@ on:
 
   schedule:
   - cron: '0 * * * *'
+  pull_request:
```
ncfavier, line 7: Should this be removed?

ncfavier (approved):

.github/workflows/cron-nixpkgs.yml
ysndr: Yes
I should think of a better way to do this debugging things

PR #350, opened 2021-08-26
Update manual.toml

pinpox:


tomberek:
This seems to be a user-specific flake and not for general usage.

pinpox:
Feel free to close it then. I don't mind sharing my modules and wanted to test out the new flake feature.

garbas:
I looked closely at the flake and while there are some user specific stuff in the flake things that we index are nicely separated and made reusable.

garbas (approved):

PR #355, opened 2021-09-09
Adding type argument for sops-nix flake

ysndr:


garbas (approved):
Ups, we should probably have a GitHub action to check the format of toml files

PR #357, opened 2021-09-10
Translate example/default values from JSON to NIx

### LINKED ISSUE #351: attribute sets and lists are displayed in a wrong format
**Describe the issue**
On https://search.nixos.org/options, one can search for NixOS-options.
Unfortunately, attribute sets and lists are displayed in a wrong way:
- correct would be: `[{ key = "value"; key2 = "value2";} {...}]`
- on website/wrong: `[{ "key": "value", "key2": "value2"}, {...}]`
  (errors: quotes around keys / : instead of = / , instead of ; / no ; at end of each option / , instead of space between items)

If copy-n-pasted to the config-file, nix reports syntax errors.

**Page links**
- https://search.nixos.org/options
- e.g. https://search.nixos.org/options?channel=21.05&show=services.xserver.xrandrHeads&from=0&size=50&sort=relevance&query=xrandr, but also many others / probably all lists and attribute sets

ysndr:
Resolves #351


tfc:
Hi, i see that the code seems to do the correct thing - but then i clicked on the "Deployed on https://613b384d36dc13b8939b19b5--nixos-search.netlify.app" link, where the old bad format still exists. is that a mistake in this deployment message or in the PR?

ysndr:
This change only affects the importer which will push a nix string to elastic rather than JSON for the examples.
As it is not part of the main branch it hasn't done so yet which is why the frontend still shows the old values.

tfc (approved):

garbas:
I've bumped the version and able the import on the PR so we can test that the text is formatted correctly.

tfc:
@garbas is the latest test deployment already supposed to show this? as i can see, it doesnt, unfortunately.

garbas:
> @garbas is the latest test deployment already supposed to show this? as i can see, it doesnt, unfortunately.

Not yet. The import needs to happen first.

@ysndr We have an error with `cron-flakes.yml`

ysndr:
> @ysndr We have an error with cron-flakes.yml

@garbas That's because sops-nix's options do not evaluate causing the whole flake group to fail. #358 would change to a model that would allow flakes to fail (which they might temporarily anyway)

The error with sops is that its options refer back to a config with `users.users` set which it isn't in our case obviously..
I just dont know whether we should work around this on a per package basis or leave it to the maintainers

We can also remove sops for now to make flake-import happy

tfc:
Just had a look at the test deployment. Looks great now!

tfc:
@garbas what is the state of this, can this be merged? i do not have approval rights but i am very interested to have this fix online.

tfc:
@garbas is there anything i can do to help unblock this?

flake-info/src/data/prettyprint.rs
```diff
@@ -0,0 +1,165 @@
...
+        assert_eq!(
+            print_value(json),
+"{
+  hello = \"world\";
+}"
+        );
+    }
+
+   
+
```
samueldr, line 163: Needs formatting?

tomberek (commented):
Changes look okay to solve the problem at hand. Simpler to add a simple pretty-printer than to integrate with rnix (but that might be a good future change).

Additional work (separate PR) could be to add a `foreman start` mechanism for hacking on the project similar to what Hydra has. This would allow running elasticsearch locally. Otherwise it's tedious to push the generated artifacts to the live ES deployment to view the results in-situ.

ysndr:
@ncfavier Im looking into the string issues you raised, in the meantime can we merge this PR as it continues to cause some confusion.

ncfavier (approved):

PR #358, opened 2021-09-10
Don't fail import of groups as long as some imports succeed

ysndr:
Currently, imports of groups fail if any flake in the group fails
This PR changes to a fail silent-model where imports succeed as long as some flake succeeds.

We should have a way to notify about failing flakes nevertheless

### DIFF
--- a/flake-info/src/bin/flake-info.rs
+++ b/flake-info/src/bin/flake-info.rs
@@ -270,7 +270,7 @@ async fn run_command(
                     return Err(FlakeInfoError::Group(errors));
                 }
                 warn!("=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=");
-                warn!("Some group memebers could not be evaluated: {}", FlakeInfoError::Group(errors));
+                warn!("Some group members could not be evaluated: {}", FlakeInfoError::Group(errors));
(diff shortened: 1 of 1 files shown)


ysndr:
@garbas can we approve this? Or at least #361 for now?
With the recent schema update the flake index causes 404 replies (#362)

samueldr (approved):
At face value this looks right.

flake-info/src/bin/flake-info.rs
```diff
@@ -265,7 +265,14 @@ async fn run_command(
                 .collect::<Vec<_>>();
 
             if !errors.is_empty() {
-                return Err(FlakeInfoError::Group(errors));
+
+                if exports.is_empty() {
+                    return Err(FlakeInfoError::Group(errors));
+                }
+                warn!("=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=");
+                warn!("Some group memebers could not be evaluated: {}", FlakeInfoError::Group(errors));
```
samueldr, line 273:
```suggestion
                warn!("Some group members could not be evaluated: {}", FlakeInfoError::Group(errors));
```

Been looking at too many memes lately?

samueldr:
Ping me (maybe on Matrix even) if it ends up causing issues, and you have a fix to get merged.

PR #359, opened 2021-09-11
feat: add nix-cargo-integration and rust-nix-templater flakes

yusdacra:


ysndr:
`rust-nix-templater` doesn't evaliate on flake-info (at least not on darwin systems)
It smells a bit like [IFD](https://nixos.wiki/wiki/Import_From_Derivation) as cargo needs to be build first to produce something able to export(?).
You might look into this in general as IFD is considered bad practice.
Ragardless, I think the flake might run fine on linux, but i cannot test this

yusdacra:
Hmm, I don't have a darwin system so I never knew if it did work fine on it. On linux it should not trigger any IFD (`nix flake show` doesn't at least).

ysndr:
Tracking it down its an issue with the `app` attribute.
We do read apps as well but aren't showing them on the frontend yet

I attached the log if you want to dig into it. I believe this will be a common problem with all crates built with your tool.
```
error: --- Error ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- nix
a 'x86_64-linux' with features {} is required to build '/nix/store/qlp5h7xb72wmdil9i4kh3xkwxkzhha43-crates-io-structopt-0.3.21.drv', but I am a 'x86_64-darwin' with features {benchmark, big-parallel, nixos-test, recursive-nix}
----------------------------------------------------------------------------------------------------- show-trace -----------------------------------------------------------------------------------------------------
trace: while evaluating 'buildCrate'
at: (239:5) in file: /nix/store/n5d7v30m07pj8bafgjxdmcm8qq3gfq0d-source/utils.nix

   238|   buildCrate =
   239|     { root
      |     ^
   240|     , memberName ? null

trace: from call site
at: (152:15) in file: /nix/store/n5d7v30m07pj8bafgjxdmcm8qq3gfq0d-source/build.nix

   151|       let
   152|         pkg = lib.buildCrate
      |               ^
   153|           (

trace: while evaluating the attribute 'paths' of the derivation 'rust-nix-templater-0.2.5'
at: (205:7) in file: /nix/store/ds4jmy7y3211axnfgl05vp04aw692yfi-source/pkgs/stdenv/generic/make-derivation.nix

   204|     // (lib.optionalAttrs (attrs ? name || (attrs ? pname && attrs ? version)) {
   205|       name =
      |       ^
   206|         let

trace: while evaluating anonymous lambda
at: (420:24) in file: /nix/store/ds4jmy7y3211axnfgl05vp04aw692yfi-source/lib/attrsets.nix

   419|     let f = attrPath:
   420|       zipAttrsWith (n: values:
      |                        ^
   421|         let here = attrPath ++ [n]; in

trace: from call site
at: (369:15) in file: /nix/store/ds4jmy7y3211axnfgl05vp04aw692yfi-source/lib/attrsets.nix

   368|       inherit name;
   369|       value = f name (catAttrs name sets);
      |               ^
   370|     }) names);

trace: while evaluating the attribute 'program'
at: (369:7) in file: /nix/store/ds4jmy7y3211axnfgl05vp04aw692yfi-source/lib/attrsets.nix

   368|       inherit name;
   369|       value = f name (catAttrs name sets);
      |       ^
   370|     }) names);
```

yusdacra:
Ah yeah i remember this now. This occurs because of crate2nix, which uses IFD. The default in nix-cargo-integration is naersk so crate2nix (and therefore this behaviour) is opt-in. I will try to work around the IFD and report back (i thought i already worked around this, but it seems i missed something or something regressed), thanks for this!

yusdacra:
I made `rust-nix-templater` use naersk in the meanwhile, so the IFD issue should be gone now.

ysndr (approved):

PR #361, opened 2021-09-16
Remove sops-nix flake

ysndr:
Sops nix fails to import


ysndr:
Nix flake search is broken currently due to a new schema version that has no flakes populated due to sops-nix failing to import

ysndr:
This PR is parallel to #358 which would ensure the flake index is populated even though some flakes fail (which they will be once we have automatic imports)

garbas (approved):

PR #363, opened 2021-09-22
Add kalker to the flakes index

Kloenk:
Kalker/kalk is a calculator with math syntax that supports user-defined variables and functions, complex numbers, and estimation of derivatives and integrals.

https://github.com/PaddiM8/kalker

### DIFF
--- a/flakes/manual.toml
+++ b/flakes/manual.toml
@@ -32,3 +32,8 @@ repo = "nickel"
 type = "github"
 owner = "pinpox"
 repo = "nixos"
+
+[[sources]]
+type = "github"
+owner = "PaddiM8"
+repo = "kalker"


ysndr:
Relevant [reddit thread](https://www.reddit.com/r/rust/comments/pskqhk/comment/hdqc4l5?context=3)

ysndr:
This flake is incompatible with the flake-info script

https://github.com/PaddiM8/kalker/blob/967b9f372b8967d1184e0796562b31d42ec0ddaf/flake.nix#L40

defines `packages` as `nixpkgs` whicle it should be an attr set of those packages build by that project (only kalker in this case)

ysndr (changes requested):
Please change the `packages` attribute for this flake

https://github.com/PaddiM8/kalker/blob/967b9f372b8967d1184e0796562b31d42ec0ddaf/flake.nix#L40

Kloenk:
My bad. Thought I did, will fix

ysndr (approved):
LGTM

Curently the frontend will only pick up derivation descriptions, therefore I recommend adding a description atrribute to the kalker derivation as well so we know what it is when it turns up in the search.

Kloenk:
> Curently the frontend will only pick up derivation descriptions, therefore I recommend adding a description atrribute to the kalker derivation as well so we know what it is when it turns up in the search.


You mean the usual meta attribute as used in nixpkgs?

PR #376, opened 2021-11-30
release: 21.05 -> 21.11

tomberek:
cc: @garbas @ysndr @jonringer @nrdxp

### DIFF
--- a/.github/workflows/cron-nixpkgs.yml
+++ b/.github/workflows/cron-nixpkgs.yml
@@ -16,6 +16,7 @@ jobs:
       matrix:
         channel:
           - unstable
+          - 21.11
           - 21.05
           - 20.09
 
--- a/src/Search.elm
+++ b/src/Search.elm
@@ -477,8 +477,8 @@ createUrl toRoute model =
 
 type Channel
     = Unstable
-    | Release_20_09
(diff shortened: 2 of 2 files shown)


garbas:
@tomberek Usually we keep channel like `20.09` around for another month before we remove it. Can you only add `21.11` without removing `20.09`? Also can you add the `21.11` channel to `.github/workflows/cron-nixpkgs.yml` as well?

@ysndr initial 21.11 channel needs to be imported manually before we merge this PR. Otherwise the `21.11` index is going to be empty until github action populates it. Do you want manually import it or should I?

garbas:
@tomberek to avoid ping/pong in comments I just fixed things myself.

ysndr:
@garbas I'm just running the initial pull should be done in a couple minutes

nrdxp:
@garbas @ysndr what is the status? We need to get this merged ASAP

ysndr (approved):

ysndr:
@nrdxp there existed an index from a previous run that failed due to an internet outage. For some reason it coulndt be deleted (i still can't though even through REST?).
Luckily nixpkgs is on a new commit now so the import succeeded and we should be good to go  now

PR #377, opened 2021-12-02
Add nixos-mailserver to the flakes index

nlewo:
Here is the project: https://gitlab.com/simple-nixos-mailserver/nixos-mailserver

### DIFF
--- a/flakes/manual.toml
+++ b/flakes/manual.toml
@@ -76,3 +76,8 @@ url = "git+https://git.sr.ht/~kerstin/sway-timetracker?ref=main"
 type = "github"
 owner = "PolyMC"
 repo = "PolyMC"
+
+[[sources]]
+type = "gitlab"
+owner = "simple-nixos-mailserver"
+repo = "nixos-mailserver"


ysndr:
Hej, I'm sorry this module definition is not compatible with flake-info as of now for two (maybe more) reasons related to option evaluation.

1. You have parts of the "implementation" in your config definition (https://gitlab.com/simple-nixos-mailserver/nixos-mailserver/-/blob/master/default.nix#L781-798).
This is: There is an evaluation of pkgs which should be happen only when writing the config i.e in `./mail-server/monit.nix`

2. Similarly you reference `cfg.fqnd` in definitions as default value while its value is not necessarily defined and has no default (same about [this](config.services.redis.bind) redis field.

Both problems mean that flake-info cannot evaluate the module. While `pkgs` could be passed into,  the evaluation, but is not canonical as far as I understand it. The issue with references to undefined cfg values can only be fixed on your end.

For both I'd assume that it is possible to move more of the "semantic" / "implementation" to the implementing config files in `./mail-server/*`

ncfavier:
For the record, I opened https://gitlab.com/simple-nixos-mailserver/nixos-mailserver/-/merge_requests/259 upstream to fix the issues here

garbas:
@nlewo @ncfavier any update on the upstream PR/MR?

nlewo:
@ncfavier sorry, i should not have postponed your NixOS mailserver MR :(

I'm now on it and it's almost merged. @garbas we will provide an update here once merged.

ncfavier (approved):
`flake-info flake gitlab:simple-nixos-mailserver/nixos-mailserver` suceeds now

PR #382, opened 2021-12-06
Print pandoc error rather than panic-ing

### LINKED ISSUE #381: flake-info: Unhelpful error message on invalid XML
When encountering invalid XML in option descriptions, `flake-info` fails with:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: exit_code: Some(44)stdout: stderr: Invalid XML:
6:1 (224)-6:8 (231): Expected end element for: Name {nameLocalName = "service", nameNamespace = Nothing, namePrefix = Nothing}, but received: EventEndElement (Name {nameLocalName = "para", nameNamespace = Nothing, namePrefix = Nothing})
', src/data/export.rs:318:43
```
Could you show which option contains the faulty description? This would greatly simplify debugging.

cc @ysndr

ysndr:
Resolves #381

### DIFF
--- a/flake-info/src/data/export.rs
+++ b/flake-info/src/data/export.rs
@@ -316,11 +316,7 @@ impl From<import::NixOption> for Derivation {
                 PandocOption::LuaFilter(man_filter),
             ]);
 
-            let result = pandoc.execute().or_else::<PandocError, _>(|err| {
-                error!("Pandoc couldnt parse documentation of '{}'", name);
-                error!("{}", err);
-                Ok(PandocOutput::ToBuffer(String::new()))
-            }).unwrap();
+            let result = pandoc.execute().expect(format!("Pandoc could not parse documentation of '{}'", name));
 
             match result {
                 PandocOutput::ToBuffer(description) => Some(description),
--- a/flake-info/src/data/export.rs
+++ b/flake-info/src/data/export.rs
@@ -316,7 +316,7 @@ impl From<import::NixOption> for Derivation {
                 PandocOption::LuaFilter(man_filter),
             ]);
 
-            let result = pandoc.execute().expect(format!("Pandoc could not parse documentation of '{}'", name));
+            let result = pandoc.execute().expect(&format!("Pandoc could not parse documentation of '{}'", name));
 
             match result {
                 PandocOutput::ToBuffer(description) => Some(description),
--- a/flake-info/src/data/export.rs
+++ b/flake-info/src/data/export.rs
@@ -4,7 +4,8 @@
 use std::path::PathBuf;
 
 use crate::data::import::NixOption;
-use pandoc::{InputFormat, InputKind, OutputFormat, OutputKind, PandocOption, PandocOutput};
+use log::error;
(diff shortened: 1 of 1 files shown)


erikarvstedt:
Thanks, this fixes #381.
Minor issue: The error message is printed twice.

erikarvstedt:
Thanks, this fixes #381.
Remaining issues:
- The command should fail with a nonzero exit code on malformed descriptions.
- The error message is printed twice.

ysndr:
> The command should fail with a nonzero exit code on malformed descriptions.

I'm not sure if a (possibly) single malformed documentation should make the whole import fail, especially that late in the process

> The error message is printed twice

It's printed by the underlying library and again by this change. We could rely just on the library log however.

flake-info/src/data/export.rs
```diff
@@ -315,7 +316,12 @@ impl From<import::NixOption> for Derivation {
                 PandocOption::LuaFilter(man_filter),
             ]);
 
-            let result = pandoc.execute().unwrap();
+            let result = pandoc.execute().or_else::<PandocError, _>(|err| {
+                error!("Pandoc couldnt parse documentation of '{}'", name);
+                error!("{}", err);
+                Ok(PandocOutput::ToBuffer(String::new()))
+            }).unwrap();
```
ysndr, line 323:
```suggestion
            let result = pandoc.execute().expect(format!("Pandoc could not parse documentation of '{}'", name));
```

This would solve @erikarvstedt's remarks

erikarvstedt:
> I'm not sure if a (possibly) single malformed documentation should make the whole import fail, especially that late in the process

If it's not possible now, we could add a `--strict` flag and later set this as the default.

I'd like to run `flake-info` as part of my (soon to be added to nix-search) flake's test suite to detect malformed option descriptions.

erikarvstedt:
Addendum: The current `main` branch fails with a nonzero status, so no `--strict` flag is needed.

ysndr:
Well, if it fails, then with non-zero, the question is just if we should be strict about misconfigurations (as before) or not
Anyway, we could go with the expect approach above for now. This would maintain the current behaviour.
Let's keep the strictness question for later.

erikarvstedt:
> if we should be strict about misconfigurations 

Yes, definitely. Otherwise malformed description go unnoticed.

`nix build github:nixos/nixos-search/fix/pandoc-error#flake_info` fails with:
```
error[E0308]: mismatched types
   --> src/data/export.rs:319:50
    |
319 |             let result = pandoc.execute().expect(format!("Pandoc could not parse documentation of '{}'", name));
    |                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `&str`, found struct `std::string::String`
    |
    = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)
```

erikarvstedt:
Everything is working fine now, thanks! You might want to squash the commits.

ysndr:
> Everything is working fine now, thanks! You might want to squash the commits.

Great, yess I squash and merge this then

PR #383, opened 2021-12-06
flake, overlay: Rename package to `flake-info`

erikarvstedt:
#### Copy of commit msg
The package now has the same name as the `app` and the command (`bin/flake-info`).

### DIFF
--- a/.github/workflows/cron-flakes.yml
+++ b/.github/workflows/cron-flakes.yml
@@ -54,7 +54,7 @@ jobs:
 
     - name: Building import_scripts
       run: |
-        nix build ./#packages.x86_64-linux.flake_info
+        nix build ./#packages.x86_64-linux.flake-info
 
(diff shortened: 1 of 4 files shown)


erikarvstedt:
Done.
The `app` flake attr is still required for `nix run`.

ysndr:
If the derivation and binary are called the same and no app is defined `nix run` will choose the derivation and run `<derivation:flake-info>/bin/flake-info`.

> If installable evaluates to a derivation, it will try to execute the program <out>/bin/<name>, where out is the primary output store path of the derivation and name is the meta.mainProgram attribute of the derivation if it exists, and otherwise the name part of the value of the name attribute of the derivation (e.g. if name is set to hello-1.10, it will run $out/bin/hello).

https://nixos.org/manual/nix/unstable/command-ref/new-cli/nix3-run.html

But enough the bike shedding ;)

PR #385, opened 2021-12-06
flake: Remove app

erikarvstedt:
This helps promoting Nix best practices.

#### Copy of commit msg
As pointed out by Yannik Sander, defining the app is redundant because the package of the same name is used automatically.


ysndr:
The possibility to have a `defaultApp` is nice still. I do not think anyone makes use of this for this project but we should still mention this is going away with the proposed change

ysndr (approved):

PR #386, opened 2021-12-06
Remove noisy prints to stderr

### LINKED ISSUE #384: `flake-info` output is printed to stderr
The output should be printed to `stdout`, so that regular output can be distinguished from error output.

I'd like to silently run `flake-info` as part of my flake's test suite, but `flake-info flake ... 2>/dev/null` suppresses all errors.

Maybe we could also add a `--check` mode, which skips all output and only checks if `flake-info` would run successfully.

ysndr:
Flake info and dumps were printed to stderr.
These are now sent as low level logs using `log::trace` and can be optionally enabled

Also bundles pandoc and pandoc variables now for less unrelated errors in nix devShells

Resolves #384

### DIFF
--- a/flake-info/default.nix
+++ b/flake-info/default.nix
@@ -1,5 +1,5 @@
 { pkgs ? import <nixpkgs> { } }: with pkgs;
-rustPlatform.buildRustPackage {
+rustPlatform.buildRustPackage rec {
   name = "flake-info";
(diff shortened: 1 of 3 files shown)


erikarvstedt (approved):
Fixes #384.

erikarvstedt (approved):

Issue #391, opened 2021-12-17
Be more resilient to nixpkgs errors


ncfavier:
It's not the first time I see the import fail because e.g. someone put a string in `meta.maintainers` (cf. https://github.com/NixOS/nixos-search/issues/389). It would be nice to be more resilient to such failures, for example:

- in the particular case of maintainers, allow parsing a string as an entry `{name: name, github: null, email: null}`
- in general, don't fail the whole import if a single package or option fails to deserialise

piegamesde:
I think we should fix this the other way around: putting a string into `meta.maintainers` should give an evaluation error in `nixpkgs`.

sternenseemann:
This should already be inforced by ofborg in nixpkgs via `check-meta.nix` which restricts the `maintainers` field to `listOf (attrsOf str)`.

ncfavier:
Ah but `(listOf (attrsOf str)).check [ "foo" ]` evaluates to `true`... Apparently [`listOf`](https://github.com/NixOS/nixpkgs/blob/59bfda72480496f32787cec8c557182738b1bd3f/lib/types.nix#L369) and [`attrsOf`](https://github.com/NixOS/nixpkgs/blob/59bfda72480496f32787cec8c557182738b1bd3f/lib/types.nix#L398) don't bother to check the types of their elements?

sternenseemann:
Right, that's weird, but maybe something we can fix upstream, although I fear it may be quite the performance penalty for the already slow module system (the `Of` part seems to be mostly used to choose the proper merging behavior).

ncfavier:
The current import failure is caused by a use of `literalDocBook` in an option `description` (https://github.com/NixOS/nixpkgs/pull/155928).

garbas:
I think we should look again how are we importing flakes and nixpkgs.

How it works right now is quite easy to miss if some flakes are failing to import. With nixpkgs is a little better (we at least see a failed github action), but we are not really reporting it somewhere.

We need somehow to put things that fail into import into quarantine and report them to our matrix channel. And for things that are in quarantine we should use data from last import.

Issue #389, opened 2021-12-14
Search action no longer runs successfully


Mic92:
See github action error: https://github.com/NixOS/nixos-search/runs/4521001235?check_suite_focus=true

ysndr:
Thanks, I am on it

ysndr:
It is because the maintainer of `gotypist` did not add themselves to the maintainers.nix file but instead wrote is name as a string, which is unexpected to nix-flakes

pb-:
@ysndr Sorry about that, I didn't realize that a string was unacceptable :see_no_evil: . Thank you for fixing it.

ysndr:
Thats alright, its been overlooked during review everything is alright.
The action wont succeed until there is an updated nixos-unstable with that change included, lets see when that is going to land.

ysndr:
Worked around in #393

PR #393, opened 2021-12-17
Allow plain strings for nixpkgs maintainer fields

ysndr:
Implements first point as mentioned in #391 to prevent issues with plain maintainer names (#389) from happening

cc. @garbas @ncfavier

### DIFF
--- a/flake-info/src/data/export.rs
+++ b/flake-info/src/data/export.rs
@@ -216,10 +216,13 @@ impl From<import::NixpkgsEntry> for Derivation {
                     .map(|l: &License| l.fullName.to_owned())
                     .collect();
 
-                let package_maintainers = package
+                let package_maintainers: Vec<Maintainer> = package
                     .meta
                     .maintainers
-                    .map_or(Default::default(), Flatten::flatten);
+                    .map_or(Default::default(), Flatten::flatten)
+                    .into_iter()
+                    .map(Into::into)
+                    .collect();
 
                 let package_maintainers_set = package_maintainers
                     .iter()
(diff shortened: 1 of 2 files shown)


ncfavier (approved):

samueldr:
Is this a symptom of something missing in the Flakes interface?

Or is the flakes indexing assuming too much from the structure of packages?

ncfavier:
This has nothing to do with flakes a priori.

ysndr:
Yes it’s not a flakes problem
It was assuming nixpkgs metadata to be structured correctly which isn’t enforced in nixpkgs (or can easily slip a review as in this case.

samueldr:
Yes this has to do with Flakes. If we're indexing flake outputs, it means there is a de-facto interface we *can* index.

If there is no interface defined, we shouldn't index the data.

ysndr:
Coorect me if I'm wrong, but I think flakes dont have the meta structure in the first place. At least I'm deriving everything i did get from nixpkgs from the flake

samueldr:
I'm really not using Flakes, so I really don't know.

I assume you're probably right, that there is no `meta` interface defcribed in Flakes. Thus the fact that we are indexing them is *probably problematic*, as it makes it a de-facto interface not defined in Flakes. I suspect this should be circled back to the Flakes authors. (@edolstra and who else?)

ysndr:
well were not exactly indexing meta as is, but some information (license, maintainers, systems) that nixpkgs chose to wrap on a derivation basis.
with flakes this creates a bit of redundancy and little additional benefit to the user as that information might already be encoded by the flake author and system attribute

PR #395, opened 2021-12-20
Fix/explicit printing

ysndr:
`literal*` values for option examples and other doc elements caused some confusion and false prints due to serialization to and from and to JSON.
We now handle all the unpacking on the Rust fixing this issue
Also improves DocBook rendering performance, by heuristically skipping non-docbook texts


ysndr:
@ncfavier 
I see you cleaned up many options a few months ago.
Thank you for that
In fact [this change](https://github.com/NixOS/nixpkgs/blame/9ab7d12287ced0e1b4c03b61c781901f178d9d77/nixos/modules/system/boot/loader/grub/grub.nix#L486) fixes an issue with the new handling of literal values I'm working on.

Do you have an opinion whether we should backport this to the other indexed channels or use a work around on the flake-info side?

ncfavier:
Ideally it should be backported but I don't have the energy to do that right now. We should definitely resilient though, and accept invalid uses of `literal*`.

ysndr:
Yes I think the same
No need to backport these now, I'm on the resilience path now!
Enjoy the holidays!

flake-info/src/commands/flake_info.nix
```diff
@@ -82,20 +82,22 @@ let
...
+            {__type = "set"; __value = lib.mapAttrs (_:substFunction )  x; }
           else if builtins.isList x then
-            map substFunction x
+            { __type = "list"; __value = map substFunction x; }
           else if lib.isFunction x then
-            "<function>"
+            { __type = "function"; }
           else
-            x;
+            { __type = "value" ; __value = x; };
```
ncfavier, line 96: Do we really need all this? My idea was to just have `substFunction` do what it's supposed to do (i.e. replace functions with `<function>`), and do the handling of `_type`d attrsets in Rust.

ysndr: either way works i guess, but you are right its probably more performant and easier to do in rust.
However that is not the main problem here.
Its more the fact that "literalExpressions" while supposed to be strings (of a string - the literal value) turn out to be all kinds of stuff and therefore make it hard to reserialize them nicely

ncfavier:
Looks great! Two final remarks:

- I don't think this should apply to `option_type` as option types are always strings currently.
- `literalDocbook`s aren't rendered properly, see [`documentation.man.manualPages`](https://61c5f09cfa1ce4b942596b58--nixos-search.netlify.app/options?channel=unstable&show=documentation.man.manualPages&from=0&size=50&sort=relevance&type=packages&query=manualpages) for example. Probably some magic involving `showHtml` is required, like [here](https://github.com/NixOS/nixos-search/blob/a4d84213120c701fa5e5cb6841cc12082ba86e0d/src/Page/Options.elm#L215-L219).

ysndr:
> I don't think this should apply to option_type as option types are always strings currently.
Yes might be better just use the string, it was special cased before but I don't know whether for a good reason

>literalDocbooks aren't rendered properly, see documentation.man.manualPages for example. Probably some magic involving showHtml is required, like here.

this one is a bit annoying as this requires the backend to generate html in all cases, either docbook derived or as a wrapper around code blocks as the frontend cant distinguish anymore between rendering text or html

ncfavier:
Yep.

As an alternative, we could have the backend always provide DocBook (either literal or code wrapped in a `<programlisting>`) and the frontend do the rendering. I don't know if that's even worth considering.

ysndr:
Yes we have to distinguish them somehow either on the backend or giving a hint to the frontend

ncfavier (approved):
Looks good

PR #396, opened 2021-12-24
Correct pretty printing for attribute names with special characters

### LINKED ISSUE #392: Attribute keys are not quoted properly
Just tracking the issue mentioned [here](https://github.com/NixOS/nixos-search/pull/357#issuecomment-936935297): attribute keys in examples/defaults containing spaces are not quoted properly (e.g. [services.netatalk.settings](https://search.nixos.org/options?channel=unstable&show=services.netatalk.settings&from=0&size=50&sort=alpha_asc&type=packages&query=services.netatalk.settings)).

ysndr:
Attributes like
```
{
  "need some space" = true;
}
```

should now be printed correctly (before `need some space` printed without `"`)

Closes #392


ncfavier:
This doesn't seem to be solving the problem yet for [`services.netatalk.settings`](https://61c5fd6a338fb038605ad7d1--nixos-search.netlify.app/options?channel=unstable&show=services.netatalk.settings&from=0&size=50&sort=relevance&type=packages&query=netatalk.settings).

Note that you need to be a little careful about how you escape Nix strings, e.g. `$` should be handled specially. See [`escapeNixString`](https://github.com/NixOS/nixpkgs/blob/c16baf2d77dd30ce20cd2e28591613d7c9cf3fd9/lib/strings.nix#L338) in the nixpkgs lib for reference (basically just escape as JSON and replace `$` with `\$`).

ysndr:
Eh it's not pushed to elastic yet anyway so the results on the front end are not updated

ncfavier:
Ok, then you just need to handle `$`

PR #397, opened 2021-12-25
Better naming for flake attributes

### LINKED ISSUE #366: Confusing flake name/URI display
Existing display of flake identification can be confusing and difficult to copy-paste. There is also a potential of misunderstanding if the flake is official NixOS/nixpkgs as we don't control the project name.
`nixos -> wezterm-nightly`

Suggestion, utilize flake URI as-is. This is easier to understand, reinforces flake URI syntax, and can be copy-pasted. Non-ascii arrow gets in the way.
`github:pinpox/nixos#wezterm-nightly`

ysndr:
Closes #366

@tomberek is that what you had in mind?

### DIFF
--- a/src/Page/Options.elm
+++ b/src/Page/Options.elm
@@ -174,6 +174,7 @@ viewResultItem channel _ show item =
             case Html.Parser.run <| String.trim value of
                 Ok [ Html.Parser.Element "rendered-docbook" _ nodes ] ->
                     Just <| Html.Parser.Util.toVirtualDom nodes
+
                 Ok _ ->
                     Nothing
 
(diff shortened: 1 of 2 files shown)


tomberek:
Yes, exactly.

samueldr (approved):
While it's a mouthful to read (eyeful?), it's much better than the previous version. This properly shows the actual origin of the flake attributes.

tomberek (approved):

ysndr:
Yes especially options get a bit messy but I guess its fine
I am hesitant about nixpkgs though, as that would require the ref to be included too which would turn out to be `github:nixos/nixpkgs/nixos-unstable#python3Packages.locationsharinglib` which i find a bit too much for the title there

samueldr:
> I am hesitant about nixpkgs though

I don't follow. Were you thinking about changing the *Packages* tab listing to use the same syntax?

If so, I think that would be a mistake. The reason it's needed in the *Flakes* tab is because Flakes have different origins, and they need to be correctly identified. The *Packages* tab only has packages for Nixpkgs.

ysndr:
> The Packages tab only has packages for Nixpkgs.
exactly, and the channel is selected explicitly anyway

PR #400, opened 2021-12-27
Render DocBook strings containing < but not </

ncfavier:
The current check is too restrictive and will not match strings containing only self-closing tags, for example the description of [`time.timeZone`](https://search.nixos.org/options?channel=unstable&show=time.timeZone&from=0&size=50&sort=alpha_asc&type=packages&query=time.timeZone).

This change still only renders about 20% of the NixOS option descriptions. I think it's now complete, i.e. there are no strings not containing `<` that would be substantially changed by the DocBook-to-HTML conversion; I could be wrong though.


ysndr (approved):
Self closing tags… of course.
Thank you

PR #401, opened 2021-12-28
Support flake attr `nixosModule`

erikarvstedt:
Support extracting module options defined via flake attr `nixosModule` (see [nix/flake.cc](https://github.com/NixOS/nix/blob/0e90b13ab1df925e549b5d55853b65911b4b40d3/src/nix/flake.cc#L582)).
These options are currently ignored by nixos-search.

`nixosModule` options are now exported with a `flake` attr of the form `[ flake ]` (instead of `[ flake moduleName ]`, which is currently used for options defined via `nixosModules.moduleName`).

#### TODO
This change must also be reflected in the Elm and Rust layers.
Field [flake](https://github.com/NixOS/nixos-search/blob/63846c3e82d7d5f08c44a57347249b897fd44be1/flake-info/src/data/import.rs#L67) of struct `NixOption` currently consists of both a flake and a module name.
@ysndr, @ncfavier, could you have a look at this?

Here's what I'd propose for displaying the options in the web interface:
```
# Current format
github:foo/bar#mymodule -> services.bar.enable

# Possible format for options defined via `nixosModule`
github:foo/bar -> services.bar.enable
```

### DIFF
--- a/src/Page/Options.elm
+++ b/src/Page/Options.elm
@@ -68,7 +68,7 @@ type alias ResultItemSource =
     , source : Maybe String
 
     -- flake
-    , flake : Maybe ( String, String )
+    , flake : Maybe (List String)
     , flakeName : Maybe String
     , flakeDescription : Maybe String
     , flakeUrl : Maybe String
@@ -274,16 +274,19 @@ viewResultItem channel _ show item =
             Maybe.map asGithubLink item.source.source
 
         flakeOrNixpkgs =
-            case ( item.source.flakeName, item.source.flake, item.source.flakeUrl ) of
+            let
+                mkLink flake url =
+                    a [ href url ] [ text flake ]
+            in
+            case ( item.source.flake, item.source.flakeUrl ) of
                 -- its a flake
-                ( Just name, Just ( flake, attr ), Just flakeUrl_ ) ->
+                ( Just (flake :: []), Just url ) ->
                     Just
-                        [ li []
-                            [ a [ href flakeUrl_ ] [ text flake ]
-                            , text "#"
-                            , text attr
-                            ]
+                        [ li [] [ mkLink flake url ]
                         ]
+
+                ( Just (flake :: moduleName :: []), Just url ) ->
+                    Just [ li [] [ mkLink flake url, text "#", text moduleName ] ]
 
                 _ ->
                     Nothing
@@ -344,12 +347,16 @@ findSource channel source =
         flakeOrNixpkgs =
             case ( source.flake, source.flakeUrl ) of
                 -- its a flake
-                ( Just ( name, module_ ), Just flakeUrl_ ) ->
+                ( Just (name :: attrs), Just flakeUrl_ ) ->
+                    let
+                        module_ =
+                            Maybe.withDefault "(default)" <| Maybe.map (\m -> "(Module: " ++ m ++ ")") <| List.head attrs
+                    in
                     Just <|
                         List.append
                             (Maybe.withDefault [] <| Maybe.map (\sourceFile_ -> [ sourceFile_, span [] [ text " in " ] ]) sourceFile)
                             [ span [] [ text "Flake: " ]
-                            , a [ href flakeUrl_ ] [ text <| name ++ "(Module: " ++ module_ ++ ")" ]
+                            , a [ href flakeUrl_ ] [ text <| name ++ module_ ]
                             ]
 
                 ( Nothing, _ ) ->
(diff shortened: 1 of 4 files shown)


ysndr:
Thank you
I’ll look into it

ysndr:
Looks good to me in general
Im just concerned with Option duplications in case a flake does have nixosModules.moduleA and nixosModule=nixosModules.ModuleA

Regarding the rust and elm layer I don’t see much of a hurdle

erikarvstedt:
Good point. A possible fix:
If `nixosModules.moduleA` is identical to `nixosModule` (by comparing the results of `lib.optionAttrSetToDocList`), export `moduleA` as `nixosModule`.
Should I implement it like this?

ncfavier:
This is a bit messy. I don't think we should support `nixosModule` because I don't think it makes sense for a flake to export a `nixosModule` that's not in its `nixosModules`. `nixosModule` is there as a shorthand, in my opinion.

erikarvstedt:
> I don't think we should support nixosModule 

Only defining `nixosModule` is common for flakes that define only one NixOS module.
But we can indeed ignore the case where `nixosModule` is not also included in `nixosModules`.
So we could simply export `nixosModule` when `nixosModules` is empty.

ysndr:
I agree, assuming the `nodeModule` is also a key of `nodeModules` makes sense to me.
Have you tested whether the rust side complains? It should iisc as it [expects a 2-tuple](https://github.com/NixOS/nixos-search/blob/main/flake-info/src/data/import.rs#L67)

erikarvstedt:
`nixosModule` is now conditionally exported.

> Have you tested 

No, because the Rust type is obviously incompatible.

ysndr:
I did a quick adaption of the rust and elm side

erikarvstedt:
Minor nitpick:
The module path will always have either one or two components, so we should make this explicit by using types like `(str (optional str))` or `((str) or (str str)` in Rust and Elm, but only if this doesn't complicate the pattern matching and the serde importing.

ysndr:
Yes the vec solution was mostly just the quickest way but given that flakes are flat i guess its right to specialize for that

ysndr:
Ok we got a rust type now
I left the elm repr as a list for now as its easier to work with

src/Page/Options.elm
```diff
@@ -285,8 +285,8 @@ viewResultItem channel _ show item =
                         [ li [] [ mkLink flake url ]
                         ]
 
-                ( Just (flake :: attrs), Just url ) ->
-                    Just [ li [] [ mkLink flake url, text "#", text <| String.join "." attrs ] ]
+                ( Just (flake :: name :: []), Just url ) ->
```
erikarvstedt, line 288: nit: `name` -> `moduleName` is clearer.

erikarvstedt:
LGTM. Can we get a netlify demo?

ysndr:
> Can we get a netlify demo?

Apparently thats not possible on forks as they dont have accesst to the netlify credentials..

ysndr:
that being said we haven't imported any flake with `nixosModule` yet so the change would not be visible

erikarvstedt:
I've squashed the fixups.

erikarvstedt:
There's one issue remaining:
Right now, the search index only contains the part after the pound sign of each option path (e.g. `weblate#weblate -> services.weblate.enable` for `github:ngi-nix/weblate#weblate -> services.weblate.enable`).
Because for all flakes in the index the module name is identical to the flake name, it's currently possible to list all options of a flake by searching for the flake name ([Example](https://search.nixos.org/flakes?channel=unstable&from=0&size=30&sort=relevance&type=options&query=weblate)).
This will no longer be possible for flakes that only define `nixosModule`. Can we add the flake name to the search index to fix this?

ysndr:
We do in fact index the flake {name, url, description}
Yet with or without the moduleName this field hasn't been indexed at all so far and if i understand elastic right was therefore never subject to search.
Actually, none of the flake information is explicitly queried even, but elastic might take it into consideration regardless - i'm not sure about that.

erikarvstedt:
Yes, I was talking about the elastic search backend. How can we find out if it includes the flake name when searching for options?

ysndr:
it is [indexed](https://github.com/erikarvstedt/nixos-search/blob/f3571efdef93581f75f1cb1aa5d51ba48dcc592e/flake-info/src/elastic.rs#L24-L27)

but [not queried](https://github.com/erikarvstedt/nixos-search/blob/5e6b2a6f24eb9f59371c029eb64c503aee356a38/src/Page/Options.elm#L397-L413)

I assume ES only look into fields explicitly [queried](https://github.com/erikarvstedt/nixos-search/blob/70b72a343fbad863d219602f075cdacab2a451b3/src/Search.elm#L1305-L1306)

erikarvstedt:
I think it would make sense to include `flake_name` in the query, with a ranking weight lower than `option_description`.

ysndr:
> I think it would make sense to include flake_name in the query, with a ranking weight lower than option_description.

I agree but for me thats a separate PR

erikarvstedt:
Alright, we can merge this then.

ysndr (approved):

PR #402, opened 2021-12-30
Fix xrefs

ncfavier:
A temporary fix for points 1 and 2 of #303.

This is bad for many reasons:

- only works for NixOS options. I don't even know how non-NixOS options are supposed to reference one another, and this might change anyway with the markdown conversion(?).
- relies on heuristics for undoing [the conversion](https://github.com/NixOS/nixpkgs/blob/7771661d93bad2f3d1d7c65852a918afd2a2bcf1/nixos/lib/make-options-doc/options-to-docbook.xsl#L23) between option names and XML ids, which will probably break at some point or for some weird edge cases.
- hardcodes `unstable` as the channel to use in links. Otherwise we would need to pass the channel name along all the way to the `render` function, which requires some refactoring.
- creates a temporary file containing the Lua filter for each call to `render`. I don't know enough Rust to figure out how to make the file persist between calls while still being deleted on exit. Also, maybe it would be better to ship those files (`fix-xrefs.lua`, `flake_info.nix`, `packages-config.nix`) in the `flake_info` package directly instead of using temporary files.

...but it should at least fix the immediate issue.

cc @jtojnar


jtojnar:
Does not seem to work here: https://61cdc6e06b639f6c0dc31cf0--nixos-search.netlify.app/options?channel=21.11&show=services.openssh.authorizedKeysFiles&from=0&size=50&sort=relevance&type=packages&query=services.openssh.authorizedKeysFiles

ncfavier:
The backend changes aren't propagated to elasticsearch for the automated deployments. I've triggered the nixpkgs import workflow on this branch manually now.

The rendered description for that option is

```xml
<rendered-docbook><p>Specify the rules for which files to read on the host. This is an advanced option. If you're looking to configure user keys, you can generally use <a href="https://search.nixos.org/options?channel=unstable&amp;show=users.users.&lt;name&gt;.openssh.authorizedKeys.keys&amp;query=users.users.&lt;name&gt;.openssh.authorizedKeys.keys">users.users.&lt;name&gt;.openssh.authorizedKeys.keys</a> or <a href="https://search.nixos.org/options?channel=unstable&amp;show=users.users.&lt;name&gt;.openssh.authorizedKeys.keyFiles&amp;query=users.users.&lt;name&gt;.openssh.authorizedKeys.keyFiles">users.users.&lt;name&gt;.openssh.authorizedKeys.keyFiles</a>. These are paths relative to the host root file system or home directories and they are subject to certain token expansion rules. See AuthorizedKeysFile in man sshd_config for details.</p>
</rendered-docbook>
```

> <p>Specify the rules for which files to read on the host. This is an advanced option. If you're looking to configure user keys, you can generally use <a href="https://search.nixos.org/options?channel=unstable&amp;show=users.users.&lt;name&gt;.openssh.authorizedKeys.keys&amp;query=users.users.&lt;name&gt;.openssh.authorizedKeys.keys">users.users.&lt;name&gt;.openssh.authorizedKeys.keys</a> or <a href="https://search.nixos.org/options?channel=unstable&amp;show=users.users.&lt;name&gt;.openssh.authorizedKeys.keyFiles&amp;query=users.users.&lt;name&gt;.openssh.authorizedKeys.keyFiles">users.users.&lt;name&gt;.openssh.authorizedKeys.keyFiles</a>. These are paths relative to the host root file system or home directories and they are subject to certain token expansion rules. See AuthorizedKeysFile in man sshd_config for details.</p>

which looks correct. I've checked a few different cases manually myself.

ncfavier:
Just rebased on latest main containing https://github.com/NixOS/nixos-search/commit/b6f1dc9667a2505f4c50b3458ee4bd795142d9cb, which affects `services.openssh.authorizedKeysFiles`. Re-running the workflow now.

ncfavier:
Hm, I bumped the VERSION file and it seems to have successfully imported the new options but the changes still aren't visible. Not sure how this works.

ysndr:
For what I can tell it's showing correctly [here](https://61cdd57b9a1becc44741db4e--nixos-search.netlify.app/options?channel=21.11&show=services.openssh.authorizedKeysFiles&from=0&size=50&sort=relevance&type=packages&query=services.openssh.authorizedKeysFiles)

ysndr (approved):
Seems to work fine as far a s I can tell
Do we need to change something in the guess method whether to run pandoc or not?

ncfavier:
Indeed. I don't really understand how the frontend connects to the backend.

ncfavier:
> Do we need to change something in the guess method whether to run pandoc or not?

I don't think so

jtojnar:
Clicking the link changes the URL in the address bar but does not change the search until I refresh the page in Firefox. (Probably the same issue as back/forward browser buttons not working.)

ncfavier:
Opened https://github.com/NixOS/nixos-search/issues/403

PR #405, opened 2021-12-31
Use upstream nixpkgs' packages-config.nix

sternenseemann:
The files in nixos-search and nixpkgs have already unnecessarily
diverged — more than they should given that the only difference is that
the search ignores unfree packages (while nixpkgs includes them for the
benefit of repology).

Since we update the NIX_PATH when calling nix-env in nixpkgs_info.rs, we
can simply import the packages-config.nix from nixpkgs using that
information and update it with our addition `allowUnfree = false`.

cc @Atemu @andir


flake-info/src/commands/packages-config.nix
```diff
@@ -1,49 +1,4 @@
-{
-  # Ensures no aliases are in the results.
-  allowAliases = false;
-
+import <nixpkgs/pkgs/top-level/packages-config.nix> // {
   # Do *NOT* list unfree packages
   allowUnfree = false;
```
ncfavier, line 3: Isn't this the default? If we can remove this we can use the file from nixpkgs directly

sternenseemann: Right, I was a bit confused because <https://github.com/NixOS/nixos-search/pull/343> changed the value, so I thought it might be necessary.

Keeping the file that reimports from nixpkgs could still be valueable to add custom changes to the upstream one? In any case I'll update the commit message and contents later.

garbas (approved):

PR #404, opened 2021-12-31
Open internal links in a new window

ncfavier:
Temporary fix for #403

PR #406, opened 2021-12-31
Fix perform deeper equality checks for internal links to the same model

### LINKED ISSUE #403: Internal links don't cause the page to change
As mentioned in https://github.com/NixOS/nixos-search/pull/402#issuecomment-1003262899, when clicking a link in an option description to a search.nixos.org page, like [here](https://search.nixos.org/options?channel=unstable&show=environment.profileRelativeSessionVariables&from=0&size=50&sort=alpha_asc&type=packages&query=environment.profileRelativeSessionVariables), the URL bar changes but the page doesn't (in both Firefox and Chromium).

@garbas do you know how to fix this?

As a temporary workaround I suggest we add `target=_blank` to those links.

ysndr:
Fixes #403 on the Elm level instead of a workaround as in #404.
Also brings support for native browser navigation

### DIFF
--- a/src/Main.elm
+++ b/src/Main.elm
@@ -191,17 +191,17 @@ pageMatch m1 m2 =
         ( Home _, Home _ ) ->
             True
 
-        ( Packages _, Packages _ ) ->
-            True
+        ( Packages model_a, Packages model_b ) ->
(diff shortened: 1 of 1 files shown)


ncfavier (approved):
Thanks!

PR #407, opened 2021-12-31
Ignore result data

ysndr:
#406 would unnecessarily reload pages if `show=` was toggled.
The checking function compares pages but the new page will be loading while the old one has results already. Hence also ignore `resultz state.

Sorry @ncfavier to have another PR on the same issue

### DIFF
--- a/src/Main.elm
+++ b/src/Main.elm
@@ -37,6 +37,7 @@ import Search exposing (defaultFlakeId)
 import Search exposing (channels)
 import Html exposing (sup)
 import Html exposing (small)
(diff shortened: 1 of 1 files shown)


ncfavier (approved):
No problem, I'll wait a bit before merging in case you can think of anything else.

ysndr:
Yes ill have a look later and merge it myself then 👍🏼

ncfavier:
Actually I just realised how annoying this issue is 😅 merging now

ysndr:
Yes 😅
Thanks
Happy new year

PR #411, opened 2022-01-07
Also ignore installDetail state

### LINKED ISSUE #409: Clicking On non-NixOS on any package on webpage doesnt have any effect
Previously clicking "On non-NixOS" tab would show install command for non-nixos environment for any package. Seems to just reload the page now.

ysndr:
fixes #409

### DIFF
--- a/src/Main.elm
+++ b/src/Main.elm
@@ -12,7 +12,9 @@ import Html
         , header
         , img
         , li
+        , small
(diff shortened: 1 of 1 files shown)


samueldr (approved):
I can't vouch for the implementation, but it sure fixes the linked issue.

PR #415, opened 2022-01-16
Cleanup workflows

ncfavier:
The install-nix-action now installs a Flakes-enabled Nix by default. I also removed the call to jq because I think it's currently just obscuring an error.


ncfavier:
Indeed the index names were wrong. I reintroduced jq and made it print the whole response if `took` is missing.

ncfavier:
Everything looks fine now: [frontend](https://github.com/NixOS/nixos-search/runs/4832344257?check_suite_focus=true), [flakes](https://github.com/NixOS/nixos-search/runs/4832344258?check_suite_focus=true), [nixpkgs](https://github.com/NixOS/nixos-search/actions/runs/1704348247).

.github/workflows/import-nixpkgs.yml
```diff
@@ -0,0 +1,63 @@
...
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: true
+      matrix:
+        channel:
+          - unstable
+          - 21.11
+          - 21.05
+          - 20.09
```
garbas, line 21: We should probably retire `20.09` channel.

ncfavier: I was thinking the same since it's EOL

garbas (approved):
:+1:

.github/workflows/import-nixpkgs.yml
ncfavier: Probably belongs in a separate PR though

ncfavier:
I will!

PR #416, opened 2022-01-17
Drop 20.09 channel

ncfavier:
Since it has reached the end of its support lifetime and is [not being updated](https://status.nixos.org/) anymore.

### DIFF
--- a/.github/workflows/import-nixpkgs.yml
+++ b/.github/workflows/import-nixpkgs.yml
@@ -18,7 +18,6 @@ jobs:
           - unstable
           - 21.11
           - 21.05
-          - 20.09
 
     env:
       RUST_LOG: debug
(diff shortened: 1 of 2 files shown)


ncfavier:
I don't understand what this list is for, or why there's a 21.09 in there:

https://github.com/NixOS/nixos-search/blob/f31e580e52ad49035ecc016fd133004762d2e5ef/src/Search.elm#L589-L615

garbas:
> I don't understand what this list is for, or why there's a 21.09 in there:
> 
> https://github.com/NixOS/nixos-search/blob/f31e580e52ad49035ecc016fd133004762d2e5ef/src/Search.elm#L589-L615

I think this is something @ysndr added when trying to use flakes backend for nixpkgs as well. I think we should at least comment this portion out if not remove it completely.

garbas (approved):

ncfavier:
Merging now, I'll leave it to him for another PR.

ysndr:
The flakes were designed to be grouped with groups for the nixpkgs versions
Yet we decided to keep nixpkgs separate what is left is different groups in the front end code

PR #418, opened 2022-01-20
CI: don't fail-fast

ncfavier:
Import failing for a channel/group doesn't mean it will fail for the other channels/groups


garbas (approved):
I wonder if this is the behavior we want **</me is thinking out loud>**. But maybe this the best scenario for **Right Now**. For this I'm improving this PR.

I think we shouldn't fail importing all of the flakes in a group if only one fails, but rather (1) report back which one failed and (2) notify the author. Maybe we should put the failing flake on some sort of a quarantine list until issues are resolved.

PR #419, opened 2022-01-20
Show package outputs

### LINKED ISSUE #414: Show .dev sub-packages in search
**Introduce your stance**
Was trying to figure out why xcbutilkeysyms doesn't register itself in pkg-config. Thought it could be a package issue, but then discovered that "xcbutilkeysyms.dev" contains the pkg-config script. Was confused as to why https://search.nixos.org/ didn't show the dev package to me.

**Describe the issue**
Dev sub-packages are a bit of a magic. It would be great if https://search.nixos.org/ assisted in discovering them.

**Page links**
https://search.nixos.org/packages?channel=21.11&show=xorg.xcbutilkeysyms&from=0&size=50&sort=relevance&type=packages&query=xcbutilkeysyms

**Additional context**

ncfavier:
Fixes #414


ncfavier:
Fixes https://github.com/NixOS/nixos-search/issues/414

This is now waiting either on https://github.com/NixOS/nix/pull/5922#issuecomment-1021211734 or on https://github.com/NixOS/nixos-search/pull/343#issuecomment-1021287999

ncfavier:
Frontend looks good now. For reference: [nix](https://61f02ac32c7b22531a13fa96--nixos-search.netlify.app/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=nix) (2 outputs), [linux](https://61f02ac32c7b22531a13fa96--nixos-search.netlify.app/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=linux) (1 output), [stdenv](https://61f02ac32c7b22531a13fa96--nixos-search.netlify.app/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=stdenv) (0 outputs).

ncfavier:
This is ready for review (but not quite ready to merge, wait until https://github.com/NixOS/nixpkgs/pull/163672 lands in nixos-unstable).

I had to drop the 21.05 channel because it won't evaluate with `packages-config.nix` and it's EOL.

I made some improvements to the frontend, following some of the discussion in https://github.com/NixOS/nixos-search/pull/429 (cc @c0bw3b).

In the future I would like to patch Nix to add an option to query only the output *names* and not the paths. This has the double benefit that we don't need to actually evaluate everything (which currently requires adding more swap to the nixpkgs worker...), and we get the *ordered* list of outputs, so we can treat the first/main output specially. Currently the outputs are printed in no particular order, except if the only output is `out`.

### Screenshots

screenshot-2022-03-11-185458
screenshot-2022-03-11-185505

ysndr:
Looking into this tomorrow or monday.
First sight looks good.
I'm not sure if were happy dropping 21.05 already (even if its EOL we used to have 3 generations of nixpkgs)
Anyway, good work thanks for your help

ncfavier:
Added sorting so things look a bit more consistent.

flake-info/README.md
```diff
@@ -181,26 +181,23 @@ Enter the [nixos/nix](https://hub.docker.com/u/nixos/) docker image and proceed
...
 ### Installation, finally
 
 This project is defined as a flake therefore you can build the tool using
 
 ```
-$ nix build <project root>
+$ nix build <project root>#flake-info
 or
-$ nix build github:miszkur/github-search
+$ nix build github:nixos/nixos-search#flake-info
```
ysndr, line 200: Should we have a separate PR to update the readme?

.github/workflows/import-nixpkgs.yml
```diff
@@ -17,14 +17,18 @@ jobs:
...
       RUST_LOG: debug
       FI_ES_EXISTS_STRATEGY: abort
       FI_ES_URL: ${{ secrets.ELASTICSEARCH_URL }}
 
     steps:
+    - name: Increase swap space
+      uses: pierotofy/set-swap-space@v1.0
+      with:
+        swap-size-gb: 10
+
```
ysndr, line 31: Is this necessary due to this pr or could it be a seperate pr itself?

ysndr (approved):
Looks fine to me.

Since 
>  The first element of outputs determines the default output. Thus, you could also write
>```
>buildInputs = [ pkg pkg.headers ];
>```

We might want to mark that default output, especially if a derivation does not have an `$out` output

flake-info/README.md
ncfavier: Why split?

.github/workflows/import-nixpkgs.yml
ncfavier: Yes, this is necessary due to the extra evaluation caused by `--out-path`

ysndr:
granted, I read "the future is now" as if it wouldn't be a problem anymore

ncfavier:
Thanks, though we should probably have launched an import before merging to avoid rolling back. Started one now

c0bw3b:
> I made some improvements to the frontend, following some of the discussion in #429 (cc @c0bw3b).

Sorry I'm late to this party but the look and UX is 👍 
Thanks for this.

PR #432, opened 2022-02-26
Add github action to check flake groups

ysndr:
Resolves #430 (and closes #356)

PR #433, opened 2022-02-27
Use composite action to reuse setup steps

ysndr:
All our actions use the same setup.
To make updating easier I implemented a composite action that can be reused by all other actions

(also includes adaptions for the check implemented in #432)

### DIFF
--- a/.github/actions/common-setup/action.yml
+++ b/.github/actions/common-setup/action.yml
@@ -1,5 +1,8 @@
 name: Setup Environment
-
+inputs:
+  CACHIX_SIGNING_KEY: # determines what node version to install
+    required: true
+    description: 'Cachix Signing Key'
 runs:
   using: "composite"
   steps:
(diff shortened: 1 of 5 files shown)


ncfavier:
Good idea, could we also factor out the checkout step?

Also, why are we using `fetch-depth: 0`? I previously thought this made a shallow clone, but it in fact makes a deep clone, shallow is the default.

garbas:
> Good idea, could we also factor out the checkout step?
> 
> Also, why are we using `fetch-depth: 0`? I previously thought this made a shallow clone, but it in fact makes a deep clone, shallow is the default.

I think this is needed for flakes.

garbas (approved):

ncfavier:
> > Good idea, could we also factor out the checkout step?
> > Also, why are we using `fetch-depth: 0`? I previously thought this made a shallow clone, but it in fact makes a deep clone, shallow is the default.
> 
> I think this is needed for flakes.

I don't think so, I have similar flakes-based workflows on my own repos and shallow clones are fine. Will make a follow-up PR

.github/actions/common-setup/action.yml
```diff
@@ -0,0 +1,23 @@
+name: Setup Environment
+inputs:
+  CACHIX_SIGNING_KEY: # determines what node version to install
```
ncfavier, line 3: I don't understand this comment. Why not use `secrets.CACHIX_SIGNING_KEY` directly?

ysndr: You can't access secrets from the action.
Also the checkout cant be factored out into the action as the action file has to be present before gh can call it.

PR #436, opened 2022-02-28
Add group reports for flake-group check

ysndr:
flake-info will populate an environment variable with a report for all failed flakes of a group if applicable.
This should help detecting faulty flakes more easily

- adds a `--report` flag to `flake-info group` which will create a report of failed flakes if some flakes of a group fail to evaluate.
- if all flakes fail the process will throw an error as it is more likely that a major issue occured

## Example

image

### DIFF
--- a/.github/workflows/check-flake-files.yml
+++ b/.github/workflows/check-flake-files.yml
@@ -4,11 +4,10 @@ on:
   workflow_dispatch:
   pull_request:
     paths:
-      - 'flakes/**.toml'
+      - "flakes/**.toml"
 
 jobs:
   automatic-custom-flakes-check:
-
     runs-on: ubuntu-latest
 
     strategy:
--- a/.github/workflows/check-flake-files.yml
+++ b/.github/workflows/check-flake-files.yml
@@ -52,7 +52,7 @@ jobs:
             echo "$(< ./report.txt)" >> $GITHUB_ENV
             echo "EOF" >> $GITHUB_ENV
 
-            echo "::error file=$flake_group::${ env.REPORT }"
+            echo "::error file=$flake_group::${{ env.REPORT }}"
           fi
 
           echo ::endgroup::
--- a/.github/workflows/check-flake-files.yml
+++ b/.github/workflows/check-flake-files.yml
@@ -38,14 +38,14 @@ jobs:
 
         for flake_group in flakes/**/*.toml
         do
-          echo "::group::Group \"$(basename $flake_group)\""
+          echo "::group::Group \"$(basename $flake_group .toml)\""
         
          ./result/bin/flake-info group "$flake_group" "$(basename "$flake_group" .toml)" --report
 
           if [[ -f "./report.txt" ]]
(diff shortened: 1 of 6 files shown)


ncfavier:
Sadly that is not how environment variables [work](https://mywiki.wooledge.org/Environment).

We should probably rather make that report part of the output JSON when `--json` is given.

ysndr:
Eh yes 🙈 
It was more of a quick idea I ran, since i've seen this used before and didnt think too much about it.
You're right though.

ysndr:
@ncfavier what do you think?

ysndr:
Well here we do process the output (creating annotations)
@erikarvstedt advised not to print errors on stdout and stderr is already used for logging which is the reason for this here.

I could print the report to stdout which would require `--json` and `--report` to be mutually exclusive

ysndr:
It has greater visibility since the error is not mixed into a wall of other logging output

.github/workflows/check-flake-files.yml
```diff
@@ -34,7 +33,22 @@ jobs:
...
         do
-          ./result/bin/flake-info group "$flake_group" "$(basename "$flake_group" .toml)"
+          echo "::group::Group \"$(basename $flake_group)\""
+        
+         ./result/bin/flake-info group "$flake_group" "$(basename "$flake_group" .toml)" --report
+
+          if [[ -f "./report.txt" ]]
+          then
+            had_error=1
+            echo "::error file=$flake_group::$(cat ./report.txt)"
```
ncfavier, line 46:
```suggestion
            echo "::error file=$flake_group::$(< ./report.txt)"
```

.github/workflows/check-flake-files.yml
```diff
@@ -34,7 +33,22 @@ jobs:
...
+          fi
+
+          echo ::endgroup::
+
+          exit $had_error
```
ncfavier: I think you want to `exit` outside the loop, so that we get all errors?

ncfavier (approved):

.github/workflows/check-flake-files.yml
```diff
@@ -34,7 +33,29 @@ jobs:
...
+          then
+            had_error=1
+
+            # sic.:
+            # Workaround for multi line output
+            echo "REPORT<<EOF" >> $GITHUB_ENV
+            echo "$(< ./report.txt)" >> $GITHUB_ENV
+            echo "EOF" >> $GITHUB_ENV
+
+            echo "::error file=$flake_group::${ env.REPORT }"
```
ncfavier, line 55:
```suggestion
            echo "::error file=$flake_group::${{ env.REPORT }}"
```

ysndr: Tack

ysndr:
I don't know if the explicit exit status is necessary.
Its also a shame that we have to either fail or succeed since this check may fail independently from the flake modified by a PR.
Yet showing success is probably even worse.

PR #437, opened 2022-02-28
Only build frontend if frontend sources or configuration changed

ysndr:


ncfavier (approved):
Hell yes. Also we should probably move the frontend stuff into a `frontend` directory

ysndr:
`build-and-deploy` is required to merge, but won't run for all PRs anymore.
@garbas I removed that requirement for now (for the record)

ncfavier:
(I'd expect that a check that isn't triggered doesn't prevent a branch from being merged, but I haven't tested this)

PR #439, opened 2022-02-28
CI: enable globstar

ncfavier:
I forgot that using `**` in bash requires an option to be enabled.

Alternatively, do we really need to *recursively* look for `.toml` files? Otherwise we could just go with `flakes/*.toml`.

Workflow now succeeds: https://github.com/NixOS/nixos-search/runs/5360158285?check_suite_focus=true

### DIFF
--- a/.github/workflows/check-flake-files.yml
+++ b/.github/workflows/check-flake-files.yml
@@ -33,6 +33,7 @@ jobs:
 
     - name: Try importing all custom flakes
       run: |
+        shopt -s globstar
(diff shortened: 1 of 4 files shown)


ysndr:
> Alternatively, do we really need to recursively look for .toml files? Otherwise we could just go with flakes/*.toml.

for now thats enough, we had some thought about using a more nested structure in the future.

ysndr (approved):

ncfavier:
Okay, let's keep it future-proof then.

PR #443, opened 2022-03-14
Add my erosanix flake

emmanuelrosa:


ysndr:
Hej

thanks for your submission.
Be aware that your module needs some changes. 
Please check your flake with the command in the readme.

emmanuelrosa:
I found the issue. It's fixed now.

garbas:
I still get the error when running `flake-info flake github:mmanuelrosa/erosanix`. 

@emmanuelrosa Could you look at it?

emmanuelrosa:
I got a successful run of `flake-info` after fixing the issue. It looks like the checks failed due to some other repos.

ncfavier:
Note that you didn't need to change the default, only to give it a textual representation for the manual: `defaultText = lib.literalExpression "pkgs.electrs"`. (We should probably document this in the README as it seems like the most common reason that flake-info fails)

emmanuelrosa:
OK, I went ahead and made the suggested change.

I noticed that flake-info is quite picky. For example, if the flake path is ".", meaning current directory, it complains about a syntax error, which really confused me because I thought there was a problem with the Nix code.  If the path to the flake ends with a forward slash it also complains, but that error is clear.

ncfavier:
Yeah that's fixed in https://github.com/NixOS/nixos-search/pull/460

ncfavier:
Hm actually it still complains about `.` not being an absolute path. That's annoying.

ysndr:
> Hm actually it still complains about . not being an absolute path. That's annoying. Apparently a problem with nix.

Yes it needs a `./.` I think, since the argument is passed to `builtins.getFlake (toString flake);` which needs an absolute path.
however nix turns `./.` into an absolute path while failing to interpret `.`.

ysndr:
> Another solution might be to pass in the flake reference using --override-flake.

.. and model flake-info.nix as a flake?
otherwise I'm not sure where it would pick that overridden flake from?

ncfavier:
Just `builtins.getFlake "input-flake"` (see the PR)

Issue #442, opened 2022-03-14
Index all texlive.* packages


Ekleog:
AFAIU we have almost all CTAN packages packaged already. However, they don't display in the package search page.

I'm thinking it'd make sense to index all texlive.* packages and not only texlive.combined.* ones. Unfortunately, I can't find anything when I grep for haskellPackages in this repo, which does appear in the search page.

Would such a change make sense to you?

ysndr:
The file that is (going to) be used for this is defined as part of nixpkgs: https://github.com/NixOS/nixpkgs/blob/master/pkgs/top-level/packages-config.nix

You can make a PR there, once #419 is merged, we will list nixpkgs according to that file.
Texlive is huge though, I don't exactly know about possible implications of including that group.

Ekleog:
Interesting thanks! Do you happen to know the command that can be used to regenerate the index locally, so that I can check the compile time and compiled size don't blow up? (and also add it as a comment on top of the file you linked so it can be found easier thereafter)

ysndr:
Just implemented (not yet merged in #445).

You can use 
```
$ nix run github:nixos/nixos-search/cli/nixpkgs-archive-import -- --json nixpkgs-archive <path-to-archive or folder> 
```

Sadly, you wont see it on the official frontend.
you could set up your own elasticsearch locally and configure the push operation and frontend accordingly or grep the json output (I'm sorry) for it.

ncfavier:
`github:nixos/nixos-search/cli/nixpkgs-archive-import#flake-info` *

tannerlegvold:
Heads up its looks like this got [merged into main](https://github.com/NixOS/nixos-search/commit/d2ca8bb5c97a0a7c0ecdb5849a48c12e88ec19ac) two years ago. So this issue should be closed.

ncfavier:
This issue is about indexing texlive packages. Texlive packages are still not indexed.

PR #445, opened 2022-03-16
Import Nixpkgs from path

ysndr:
Add a subcommand that allows to import nixpkgs from a local repository or archive file

I wanted to have this earlier and #442 brought this back up.
To test changes to nixpkgs (i.e. the all-packages.nix) file, we need to point to an arbitrary source of nixpkgs.

What remains is a way to see changes in a (testing) frontend but that is a separate issue

USAGE:

```
$ nix run github:nixos/nixos-search/cli/nixpkgs-archive-import -- nixpkgs-archive <path-to-archive or folder> [other options]`
```

### DIFF
--- a/flake-info/src/bin/flake-info.rs
+++ b/flake-info/src/bin/flake-info.rs
@@ -43,6 +43,7 @@ struct Args {
 
 #[derive(StructOpt, Debug)]
 enum Command {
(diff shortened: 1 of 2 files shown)


ncfavier (approved):
Looks good, I've been wishing for such a thing in the past

PR #446, opened 2022-03-16
flake: set flake-info as default package

ncfavier:
Makes more sense to have the CLI tool as default package.


ysndr (approved):
I does indeed

PR #449, opened 2022-03-17
flake-info: don't allow options to evaluate disallowed packages

ncfavier:
Options shouldn't evaluate packages in documentation attributes. These are errors that should be fixed in nixpkgs, which https://github.com/NixOS/nixpkgs/pull/136909 did.

Tested import on unstable and 21.11.


ncfavier:
Also it seems like such errors will cause an evaluation error in nixpkgs when trying to build the manual, which I was not aware of, so we don't have to worry about more import failures.

garbas:
@ncfavier :heart: This was a major work you pull of in https://github.com/NixOS/nixpkgs/pull/136909. This is really awesome.

garbas (approved):

PR #450, opened 2022-03-17
Cleanup

garbas:
* `import-nixpkgs` and `import-flakes` worflows now merged under `import-to-elasticsearch`, but still ran as separate jobs. This will allow us to have one "switch to latest import" job that requires both jobs to succeed.

* elm files were formatted and not used code removed. Next step is to work on unified search experience for both flakes and nixpkgs, but this will come in next PRs

* Use `flake-utils` in `flake.nix` to hopefully simplify stuff. More nix cleanup changes should land, but I thought this would be the first step.


ncfavier (approved):
I'm glad you did this, I was planning to convert the flake to flake-utils (but probably would have done a worse job), *and* I was planning to merge the import workflows.

I think it would also be nice to factor the "Building flake-info" step out. Make a first job that builds flake-info and pushes it to cachix, and then the two other jobs can fetch it.

garbas:
> I think it would also be nice to factor the "Building flake-info" step out: make a first job that builds flake-info and pushes it to cachix, and then the two other jobs can fetch it.

:+1: 

I'll do that

ncfavier:
`RUST_LOG` should stay where it is, it's a runtime thing.

Also, maybe we should use something like https://github.com/Mic92/nix-build-uncached or https://github.com/NixOS/nix/issues/4364#issuecomment-857781612 to avoid a useless "download flake-info from cachix and do nothing with it" step.

ncfavier:
Maybe we can worry about that last thing later though, as it seems to be an unresolved issue how to do this properly.

PR #451, opened 2022-03-17
flake-info: enforce minimal nix version

ncfavier:
We need at least nix 2.7 (or 2.7pre_old_enough) for the nixpkgs import, so give a helpful error instead of letting nix fail.

It got a bit messy because nix versions aren't quite semver-compliant (there should be a dash before pre) so I had to strip the pre-suffix.

### DIFF
--- a/flake-info/src/commands/nix_check_version.rs
+++ b/flake-info/src/commands/nix_check_version.rs
@@ -1,6 +1,6 @@
 use command_run::Command;
 use log::info;
-use semver::{VersionReq, Version};
+use semver::{Version, VersionReq};
 use thiserror::Error;
 
 #[derive(Debug, Error)]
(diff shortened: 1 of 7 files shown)


ysndr:
Let's make this a general check, i.e. test this berfore running any nix command in https://github.com/NixOS/nixos-search/blob/26caeaf4883c13b1d9f3d47c922308709930346e/flake-info/src/bin/flake-info.rs#L206

I dont quite like that some features do work with some nix versions while other wont. It can lead to quite some bad UX imo

ncfavier:
OK, but I need Rust assistance: if I move this code into `run_command`, how do I handle errors? That function returns a `FlakeInfoError`, should I add a new constructor `FlakeInfo(anyhow::Error)` or something?

ysndr:
I moved it into its own module and added an error type for it
Also declare the nix version in a cargo config file

flake-info/src/commands/nix_check_version.rs
```diff
@@ -0,0 +1,42 @@
+use command_run::Command;
+use log::info;
+use semver::{Version, VersionReq};
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum NixCheckError {
+    #[error("Installed nix doesn't match version requirement: {0} (required >={1})")]
```
ncfavier, line 8:
```suggestion
    #[error("Installed nix doesn't match version requirement: {0} (required {1})")]
```

ncfavier: Wait, it's my PR 😅

ncfavier:
Looks good now, thank you

ysndr:
have you tried it? I didnt have time to find an old nix

ysndr (approved):

ncfavier:
Yes, succeeds and fails fine.

ysndr:
image

PR #452, opened 2022-03-17
Show default outputs

ncfavier:
Following https://github.com/NixOS/nix/pull/6242, we now have access to default outputs so we can single them out in the frontend.

We can also drop `--out-path`, which speeds up import quite a lot and doesn't require increasing swap space or allowing broken/unfree packages anymore.

Blocked until the next nix release (unless we're impatient and use nix from the latest master)


ncfavier:
I decided to be impatient and install `nixUnstable` (which now contains https://github.com/NixOS/nix/pull/6242) for the nixpkgs import.

Note that we can now relax the minimum Nix version to 2.4 since we're not using `--out-path`, and indexing outputs is optional.

The 21.11 channel doesn't have outputs for now because I messed up the first import and I don't want to bump VERSION again, so maybe wait until the next nixos-21.11 update.

### Screenshots

screenshot-2022-03-25-112035
screenshot-2022-03-25-112411
screenshot-2022-03-25-112054

.github/workflows/import-to-elasticsearch.yml
```diff
@@ -58,6 +53,11 @@ jobs:
       with:
         CACHIX_SIGNING_KEY: ${{ secrets.CACHIX_SIGNING_KEY }}
 
+    - name: Installing Nix unstable
+      run: |
+        nix-env -f '<nixpkgs>' -iA nixVersions.unstable # for https://github.com/NixOS/nix/pull/6242
+        echo "$HOME/.nix-profile/bin" >> "$GITHUB_PATH"
+
```
ysndr, line 60: Why not add this to the `setup` action?
You can provision specific nix versions with `cachix/setupNix`

flake-info/src/commands/nixpkgs_info.rs
```diff
@@ -22,18 +22,9 @@ pub fn get_nixpkgs_info<T: AsRef<str> + Display>(nixpkgs_channel: T) -> Result<V
...
-        "--out-path",
         "--json",
     ]);
 
-    // Nix might fail to evaluate some disallowed packages
-    let mut env = HashMap::new();
-    env.insert("NIXPKGS_ALLOW_BROKEN".into(), "1".into());
-    env.insert("NIXPKGS_ALLOW_UNSUPPORTED_SYSTEM".into(), "1".into());
-    env.insert("NIXPKGS_ALLOW_UNFREE".into(), "1".into());
-    env.insert("NIXPKGS_ALLOW_INSECURE".into(), "1".into());
```
ysndr, line 34: Is this a result of the changes to nix?

flake-info/src/data/import.rs
```diff
@@ -125,7 +126,9 @@ pub struct Package {
     pub pname: String,
     pub version: String,
     #[serde(default)]
-    pub outputs: HashMap<String, String>,
+    pub outputs: HashMap<String, Option<String>>,
```
ysndr, line 129: Why is it `Option<String>` as value type?

ysndr (approved):
I understand most of the changes are in `nixos/Nix`

The changes here look fine to me, i just had a few small questions

.github/workflows/import-to-elasticsearch.yml
ncfavier: It's really only needed for the nixpkgs import so there's no need to fetch nixUnstable in every other job.

I know you can specify `install_url` when installing Nix but it seems clearer this way

flake-info/src/commands/nixpkgs_info.rs
ncfavier: Yes, we're not evaluating packages anymore.

flake-info/src/data/import.rs
ncfavier: See https://github.com/NixOS/nix/pull/6242#issuecomment-1066567786, we settled on making the output paths nullable.

.github/workflows/import-to-elasticsearch.yml
ysndr: Ok fair enough

ncfavier:
Started an import, will merge after it succeeds

PR #453, opened 2022-03-26
Add support for beta channels

dasJ:
Might be interesting for new releases that are already branched but not released yet.

Looks like this:
image

### DIFF
--- a/src/Search.elm
+++ b/src/Search.elm
@@ -45,6 +45,7 @@ import Html
         , h4
         , input
         , li
+        , small
         , p
         , span
         , strong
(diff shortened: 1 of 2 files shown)


ncfavier:
Can this be detected automatically? Then we could make a workflow that adds new channels and removes the beta tag appropriately.

dasJ:
Probably, yes. I don't think I can do that however because I lack skills in Workflows, Flakes and the current Tooling :(

garbas:
Lets incorporate this change into #455

PR #455, opened 2022-03-29
Make the channel importer versions dynamic

dasJ:
I have no idea if this works…

Taken from: https://thekevinwang.com/2021/09/19/github-actions-dynamic-matrix/


ysndr:
Oh ive been looking for this for a while.
I thought you couldn't use expressions at the matrix side.

garbas:
I went ahead and merge with current `main` and got first 4 items done.

ncfavier:
I think this would mean we only get new channels when we update the flake inputs? Even if that process is semi-automated, I wonder if we could update the channels truly dynamically.

garbas:
> I think this would mean we only get new channels when we update the flake inputs? Even if that process is semi-automated, I wonder if we could update the channels truly dynamically.

I would like to have control over the channels. To help us reproduce the bugs. I don't think it is a problem if channel comes online a day later.

flake.nix
```diff
@@ -26,17 +29,53 @@
...
+              builtins.map
+                (n: let v = filterChannels.${n}; in
+                  {
+                    id = pkgs.lib.removePrefix "nixos-" n;
+                    title = pkgs.lib.removePrefix "nixos-" n;
+                    status = v.status;
+                    jobset =
+                      builtins.concatStringsSep
+                        "/"
+                        (pkgs.lib.init (pkgs.lib.splitString "/" v.job));
```
ncfavier, line 54: Do you have plans to use the `jobset` in the future? Also, would `title` ever be different from `id`?

karantan:
> Would somebody be willing to work on this? I can help out if you get stuck, either via matrix or short call.

🤚 I think I can do it. Or at least give it a shot. I would also like to go on a short call if it's possible to make sure I'm going in the right direction.

garbas:
> > Would somebody be willing to work on this? I can help out if you get stuck, either via matrix or short call.
> 
> raised_back_of_hand I think I can do it. Or at least give it a shot. I would also like to go on a short call if it's possible to make sure I'm going in the right direction.

Join us here and lets see when we both have time: https://matrix.to/#/#nixos-search:matrix.org

flake.nix
garbas: > Do you have plans to use the `jobset` in the future?

Yes, showing what is in the binary cache. This used to work, but we was a little unstable and we removed it until we figured out other stuff first.

> Also, would `title` ever be different from `id`?

good question. maybe title is not needed here, but we would calculate in the frontend. I'll remove it for now.

flake-info/src/bin/flake-info.rs
```diff
@@ -251,6 +254,9 @@ async fn run_command(
             Ok((exports, ident))
         }
         Command::NixpkgsArchive { source, channel } => {
+            // TODO: if NIXOS_CHANNELS environment variable is present check
```
garbas, line 257: @karantan @ysndr  This is where step number 7 should be implemented. I've also added the instructions

ysndr:
Pushed to the wrong repo, point 7 now in here too

ysndr:
I split the dev shells for frontend and flake info in the last commit since i dont think you should need to down load 4G of elm if you only work on the rust and vice verse.
Happy to make a new PR on top of this later if that is desired.

flake.nix
```diff
@@ -10,36 +10,102 @@
...
+          devShells.flake-info = mkDevShell {
+            inputsFrom = packages.flake-info;
+            extraPackages = [ pkgs.rustfmt ];
+            extraShellHook = ''
               export RUST_SRC_PATH="${pkgs.rustPlatform.rustLibSrc}";
               export NIXPKGS_PANDOC_FILTERS_PATH="${packages.flake-info.NIXPKGS_PANDOC_FILTERS_PATH}";
             '';
           };
 
+          devShells.flake-info = mkDevShell {
```
ncfavier, line 96:
```suggestion
          devShells.frontend = mkDevShell {
```

flake-info/src/bin/flake-info.rs
```diff
@@ -404,3 +420,30 @@ async fn push_to_elastic(
 
     Ok(())
 }
+
+/// Information about allowed and default nixos channels.
+/// Typyically passed by environment variable NIXOS_CHANNELS.
+/// Used to filter the input arguments for `flake-info nixpkgs` and `flake-info nixpkgs-archive`
```
ncfavier, line 426: Not sure I understand the point of checking this in flake-info. What if I want to run `nixpkgs-archive` locally on some random branch? Will that fail?

ncfavier (changes requested):

flake.nix
```diff
@@ -10,36 +10,102 @@
...
+                    builtins.elem v.status ["beta" "stable" "rolling"] &&
+                    pkgs.lib.hasPrefix "nixos-" n &&
+                    v ? variant && v.variant == "primary"
+                  )
+                  allChannels;
+            in
+            {
+              channels =
+                builtins.map
+                  (n: let v = filteredChannels.${n}; in
```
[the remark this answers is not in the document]
ncfavier: Don't forget `(builtins.attrNames filteredChannels)` → `filteredChannels`

garbas:
> Instead of writing a JSON file, how about instead exposing the nixos channels as a `lib.nixosChannels` output (defined outside of the `eachDefaultSystems` call), and in the workflow doing something like

I plan to reuse this JSON file also for other flakes. I guess we should rename it than to something else, but I guess we can do this at the later point in time, but this way we are prepared to grow the channels (or release branches) for other things then nixpkgs.

garbas:
> I split the dev shells for frontend and flake info in the last commit since i dont think you should need to down load 4G of elm if you only work on the rust and vice verse.
> Happy to make a new PR on top of this later if that is desired.

I've reverted partially the change and made that both subprojects are there in devShell.default, but you can always select a subproject (`nix develop .#flake-info`) if you don't with to pull in the Elm stuff.

garbas (approved):

garbas:
I'm closing this PR and open new one to provide a preview for changes.

PR #457, opened 2022-03-30
Fix display of `null` and missing defaults and examples

ncfavier:
Currently `null`s are being treated as missing fields, and hence displayed as "Not given" on the frontend.

Solution adapted from https://stackoverflow.com/a/44331646, there is apparently no better way.

Diffed the flake-info output, this only changes some `null`s into `"null"`s.

Example option with a `null` default: [`boot.loader.grub.fontSize`](https://6244b544b98d1a00632e13a2--nixos-search.netlify.app/options?channel=unstable&show=boot.loader.grub.fontSize&from=0&size=50&sort=relevance&type=packages&query=grub.fontsize).
Example option with a non-given default: [`boot.binfmt.registrations.<name>.interpreter`](https://6244b544b98d1a00632e13a2--nixos-search.netlify.app/options?channel=unstable&show=boot.binfmt.registrations.%3Cname%3E.interpreter&from=0&size=50&sort=relevance&type=packages&query=boot.binfmt.registrations.%3Cname%3E.interpreter).


ncfavier:
I changed the frontend a bit to omit missing fields instead of saying "Not given"

garbas (approved):
:heart:

PR #459, opened 2022-03-31
Some frontend cleanup

garbas:
* remove Debug.log statements
* remove View.Components module (merge it into the rest of the Search module)

This work is all preparation for work where I hope to combine the Package/Options/Flakes routes into one view.


frontend/src/Search.elm
```diff
@@ -871,34 +907,91 @@ viewNoResults categoryName =
...
     -> Maybe String
     -> Html c
 viewSearchInput outMsg categoryName selectedChannel searchQuery =
     form
         [ onSubmit (outMsg QueryInputSubmit)
         , class "search-input"
         ]
-        [ div []
+        ([]
+          |> List.append
```
ncfavier, line 972: I'd like to move away from the `|> List.append` idiom as it inverts the order of the lists you're concatenating (and in this case breaks the frontend because the search input ends up below the channel selector).

Can you use `++` instead?

ncfavier (changes requested):
LGTM apart from the one breakage, and this doesn't conflict with my changes in Options.elm :)

frontend/src/Search.elm
ncfavier: Is there a reason to wrap things in `div []`?

Issue #447, opened 2022-03-16
Show renamed options properly


ncfavier:
See https://github.com/NixOS/nixpkgs/issues/164453: options like `networking.interfaces.<name>.ipAddress` are displayed (poorly) in nixos-search, while they are absent from the nixos manual.

We should either remove them, or make it clear that the option was renamed and link to the new option.

garbas:
:+1: we should definitely show that an option was renamed and provide a link to new option.

@ncfavier Is this information exposed in our export json?

ncfavier:
No, this would require changes to the nixpkgs module system to tag renamed options with the old option's name.

I probably won't have time to work on that before a while.

garbas:
Np. Lets keep it in our backlog

PR #460, opened 2022-03-31
Rely on upstream for extracting nixpkgs options

ncfavier:
Instead of extracting options from nixpkgs ourselves in `flake_info.nix`, use the generated `options.json`. As a result:

- Invisible options are not imported anymore. This includes removed/aliased/etc options, which proved to be confusing for some people: https://github.com/NixOS/nixpkgs/issues/164453. Note that there's some discussion in that issue about making those options visible in the manual and clearly indicating that they are removed/aliased/etc, but this is currently not the case and I think we should align with the nixos manual.
- Some additional virtualisation-related options like [`virtualisation.diskSize`](https://624f616336bfb7009275a3d5--nixos-search.netlify.app/options?channel=unstable&show=virtualisation.diskSize&from=0&size=50&sort=relevance&type=packages&query=virtualisation.diskSize) are now imported because of https://github.com/NixOS/nixpkgs/blob/fe9906cb4896d70d86452d76d11838025ffd6532/nixos/modules/misc/documentation.nix#L11.

Also, we can make the `flake` argument to `flake_info.nix` optional, which fixes `nixpkgs-archive`'s option import for local paths, and use `--argstr flake` instead of `--arg flake` so that we can pass flakes without a colon (e.g. flakes from the registry or paths).

Related to https://github.com/NixOS/nixos-search/issues/447


ncfavier:
OK, by using `--override-flake` we get the full convenience of Nix's supported CLI flake references. It's unfortunate that we still have to pass the flake ref as an argument for the module path; we could probably do without it as the flake ref is included in the export anyway in `flake_source.url`. I'll look into it later.

ysndr (approved):
That looks like quite a drop in replacement for nixpkgs.
The benefits are good in the sense of nixpkgs the only thing we need to watch for is that the output format does not diverge from reading other flakes' options once we do expose those on the frontend.

If it imports I'm fine with this

ncfavier:
Starting an import now

PR #465, opened 2022-04-07
Move hardcoded nix channels to Channels.elm

karantan:
This way it will be easier to update them because all the information is in one JSON blob (inside Channels.elm file).

Ref: https://github.com/NixOS/nixos-search/pull/455

### DIFF
--- a/frontend/src/Channels.elm
+++ b/frontend/src/Channels.elm
@@ -1,0 +1,17 @@
+module Channels exposing (channelsJSON)
+
+
+channelsJSON : String
+channelsJSON =
+    """
+[
+    {
+        "channel": "21.11",
+        "status": "stable"
+    },
+    {
+        "channel": "unstable",
+        "status": "rolling"
+    }
+]
+"""
--- a/frontend/src/Search.elm
+++ b/frontend/src/Search.elm
@@ -37,6 +37,7 @@ import Base64
 import Browser.Dom
 import Browser.Events exposing (Visibility(..))
 import Browser.Navigation
+import Channels
 import Html
     exposing
         ( Html
@@ -472,7 +473,7 @@ createUrl toRoute model =
 
 type Channel
     = Unstable
-    | Release_21_11
+    | Release String
 
 
 {-| TODO: we should consider using more dynamic approach here
@@ -488,7 +489,13 @@ type alias ChannelDetails =
 
 defaultChannel : String
 defaultChannel =
-    "21.11"
+    -- default channel is the first one in `channels` list
+    case channels of
+        [] ->
+            ""
+
+        default :: _ ->
+            default
 
 
 channelDetails : Channel -> ChannelDetails
@@ -497,8 +504,8 @@ channelDetails channel =
         Unstable ->
             ChannelDetails "unstable" "unstable" "nixos/trunk-combined" "nixos-unstable"
 
-        Release_21_11 ->
-            ChannelDetails "21.11" "21.11" "nixos/release-21.11" "nixos-21.11"
+        Release rls ->
+            ChannelDetails rls rls ("nixos/release-" ++ rls) ("nixos-" ++ rls)
 
 
 channelFromId : String -> Maybe Channel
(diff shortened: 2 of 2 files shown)


dasJ (commented):
The change looks like a great leap in a direction I want, but I cannot really judge the quality of the Elm code due to my lack of Elm knowledge

frontend/src/Channels.elm
```diff
@@ -0,0 +1,17 @@
...
+    {
+        "channel": "21.11",
+        "status": "stable"
+    },
+    {
+        "channel": "unstable",
+        "status": "rolling"
+    }
+]
+"""
```
ncfavier, line 17: It would be better to make this a JSON file and embed it in the Elm code somehow, if we want to be able to update it easily from outside.

frontend/src/Search.elm
```diff
@@ -472,7 +473,7 @@ createUrl toRoute model =
 
 type Channel
     = Unstable
-    | Release_21_11
+    | Release String
```
ncfavier, line 476: Why do we still have an `Unstable` constructor if the unstable channel is in Channels.elm?

frontend/src/Search.elm
```diff
@@ -488,7 +489,16 @@ type alias ChannelDetails =
...
 defaultChannel : String
 defaultChannel =
-    "21.11"
+    -- default channel is the first one in `channels` list
+    case channels of
+        [] ->
+            ""
+
+        [ default ] ->
+            default
```
ncfavier, line 498: Redundant pattern

frontend/src/Search.elm
```diff
@@ -520,11 +527,62 @@ channelDetailsFromId channel_id =
...
+        |> Json.Decode.Pipeline.required "status" Json.Decode.string
+
+
 channels : List String
 channels =
-    [ "21.11"
-    , "unstable"
-    ]
+    let
+        availableChannles =
```
ncfavier, line 551: Typo

frontend/src/Search.elm
```diff
@@ -520,11 +527,62 @@ channelDetailsFromId channel_id =
...
+        Ok chans ->
+            let
+                stableChans =
+                    List.filterMap isStable chans
+
+                unstableChans =
+                    List.filterMap isUnstable chans
+            in
+            -- order channels by status (stable first)
+            List.map (\c -> c.channel) (stableChans ++ unstableChans)
```
ncfavier, line 567: How about `List.sortBy (\c -> if c.status == "stable" then 0 else 1) chans`?

ncfavier (changes requested):

frontend/src/Channels.elm
ncfavier: For example using https://github.com/emmabastas/elm-embed

frontend/src/Search.elm
karantan: If it's `Unstable` then ChannelDetails are `ChannelDetails "unstable" "unstable" "nixos/trunk-combined" "nixos-unstable"` instead of `ChannelDetails rls rls ("nixos/release-" ++ rls) ("nixos-" ++ rls)`

ncfavier: So we don't want the unstable channel in the channel list then?

frontend/src/Channels.elm
ncfavier: This probably adds more complexity than it solves though

karantan: Can we do this in the next step? Because we need to modify the deployment (see https://github.com/emmabastas/elm-embed#linux) and it might not be as trivial

frontend/src/Search.elm
```diff
@@ -1024,16 +1067,7 @@ viewChannels outMsg selectedChannel =
...
-            []
-
-         else
-            [ p [ class "alert alert-error" ]
-                [ h4 [] [ text "Wrong channel selected!" ]
-                , text <| "Please select one of the channels above!"
-                ]
-            ]
-        )
+        []
```
ncfavier, line 1070: Please remove the now useless `List.append` here (and while you're at it, the wrapping `[ div [] ... ]` as there's already one).

ncfavier (approved):

garbas (changes requested):
I documented in #455 what would be the correct way to handle dynamic channels.

PR #473, opened 2022-04-20
Feat/dynamic import

garbas:
was #455


garbas:
And finally a successful import https://github.com/NixOS/nixos-search/actions/runs/2216914043

Issue #434, opened 2022-02-27
Make deployments less noisy


ysndr:
Each netlify deployment currently comments on both the commit as Wel as the PR aside from also leveraging deployments.
I think we should either enable obmverriding the PR comment or disable comments altogether since github already provides links to the deployment in a less intrusive way.

garbas:
If I'm not mistaking this was fixed now. Please reopen if I'm wrong.

ncfavier:
It's still very noisy, and I get notifications for commits like https://github.com/NixOS/nixos-search/commit/3df8b7b165e8ee1a7c5957153e049a36f7c9c7f0#commitcomment-71737881 that aren't even on a branch. I am in favour of overriding PR comments, and disabling commenting on commits.

PR #474, opened 2022-04-20
Make the Netlify bot quieter

ncfavier:
Fixes https://github.com/NixOS/nixos-search/issues/434 maybe

My mailbox be like


garbas (approved):

garbas:
I merged it and lets see if this makes it less noisy, but still gives us a way to see previews for each commit that was build and deployed.

PR #484, opened 2022-05-25
Add the 22.05 channel by updating the flake

dasJ:


ncfavier:
Started an import on this branch.

ncfavier:
It's trying to import `nixos-22.05` but that branch doesn't exist yet.

dasJ:
Thanks, I will ping here again when Hydra pushed the channel

dasJ:
Channel should be available now

ncfavier (approved):
Looks OK

ysndr:
Since the 22.05 becomes stable on Monday (presumably) should we wait for stable and merge then?

dasJ:
I actually think it would be better to merge this as-is. The reason is that the channel is clearly marked as Beta so it shouldn't matter as much and having the channel already available in the UI allows me to release 22.05 without having to wait for the PR to be merged in the configurations repo and then wait for the PR here. Telling users "it will still say beta although it's already released" is preferable to having to say "there is no search at all" imo. But that's my opinion, it's yours to decide

ncfavier:
Fine with me.

PR #489, opened 2022-06-06
Display deprecated channels

ncfavier:
Display `deprecated` (but not `unmaintained`) channels. Currently this includes 21.11 but not 21.05.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -12,7 +12,7 @@
   outputs = { self
             , nixpkgs
             , flake-utils
-            , nixos-org-configurations 
+            , nixos-org-configurations
             }:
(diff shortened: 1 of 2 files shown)


ysndr:
Looks fine to me, better if we could display the status of channels in the frontend somehow

ncfavier:
Yeah I'm thinking of copying the badges from https://status.nixos.org/.

ncfavier:
Done

dasJ:
Do we want the "Stable" and "Rolling" badges as well? Looks kind of strange imo (although that's probably only because I'm not used to it)

PR #500, opened 2022-07-20
Add microvm.nix to the flakes index

astro:


ysndr:
I think this flake was forgotten about, I'm sorry

For the record this is the flake: https://github.com/astro/microvm.nix

astro:
How can I reproduce the issue? `nix flake info` and `nix flake show` work.

ysndr:
```
nix run github:nixos/nixos-search#flake-info -- flake github:astro/microvm.nix
```

its we evaluate a few more fields from you package derivations than nix flake show does, so make sure your attributes (esp meta) are not defined recursively

ncfavier:
The infinite recursion is caused by the use of `modulesPath`, which we do not provide during evaluation. Maybe we should?

There are other issues once that is fixed: `microvm.bootDisk`, `microvm.socket` and `microvm.declaredRunner` all need a `defaultText` field.

ncfavier:
https://github.com/NixOS/nixos-search/pull/536

astro:
Fixed in https://github.com/astro/microvm.nix/commit/cf98e0276537977f78bfad155345e9caba4f6688

ncfavier (approved):
Looks good now

PR #501, opened 2022-07-21
Don't panic during export

ncfavier:
Avoid panicking if export fails for a single flake, e.g. if Pandoc fails to render the description. (Flake import is currently stuck because of nix-bitcoin...)

Also make sure that `get_flake_info` is only called once per flake.

### DIFF
--- a/flake-info/src/data/export.rs
+++ b/flake-info/src/data/export.rs
@@ -1,9 +1,10 @@
 /// This module defines the unified putput format as expected by the elastic search
 /// Additionally, we implement converseions from the two possible input formats, i.e.
 /// Flakes, or Nixpkgs.
-use std::{convert::TryInto, path::PathBuf};
+use std::{convert::{TryInto, TryFrom}, path::PathBuf};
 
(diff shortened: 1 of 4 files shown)


flake-info/src/data/export.rs
```diff
@@ -230,8 +241,7 @@ impl From<import::NixpkgsEntry> for Derivation {
 
                 let package_maintainers_set = package_maintainers
                     .iter()
-                    .filter(|m| m.name.is_some())
-                    .map(|m| m.name.to_owned().unwrap())
+                    .flat_map(|m| m.name.to_owned())
```
ysndr, line 244: Oh `flat_map` unboxes `Options`? Thats neat

ysndr (approved):
Looks good too me
Thanks for cleaning things up here!

flake-info/src/data/export.rs
ncfavier: Yeah, both Option and Result are iterators over their (possibly absent) success value

PR #502, opened 2022-07-21
Disallow Import From Derivation in flakes

ncfavier:
I think it is surprising that `flake-info` tries to build things in order to index flakes. Hence, forbid IFD.

`--read-only` is for performance, see https://github.com/NixOS/nix/pull/6323

Ping @srid since your flake (emanote) is currently the only one that uses IFD (but note that it fails even with IFD: `error: a 'aarch64-darwin' with features {} is required to build ...`, so it's not indexed currently)

Ping @erikarvstedt because why not

### DIFF
--- a/flake-info/src/commands/nix_flake_attrs.rs
+++ b/flake-info/src/commands/nix_flake_attrs.rs
@@ -9,7 +9,7 @@ use std::io::Write;
 use std::path::PathBuf;
 
 const SCRIPT: &str = include_str!("flake_info.nix");
-const ARGS: [&str; 3] = ["eval", "--json", "--no-write-lock-file"];
+const ARGS: [&str; 5] = ["eval", "--json", "--no-allow-import-from-derivation", "--read-only", "--no-write-lock-file"];
 
 /// Uses `nix` to fetch the provided flake and read general information
 /// about it using `nix flake info`


erikarvstedt (approved):
Forbidding IFD is sensible and the implementation is correct.

ysndr (approved):
Agreed, the change makes sense and should improve performance
Ive been using `builtins.unsafeDiscardStringContext` unaware of `--read-only`

ncfavier:
Ah, actually @yusdacra your flakes also use IFD and are now [failing](https://github.com/NixOS/nixos-search/runs/7447484123?check_suite_focus=true#step:4:95). Any idea where `discoverers` might be coming from?

yusdacra:
> Ah, actually @yusdacra your flakes also use IFD and are now [failing](https://github.com/NixOS/nixos-search/runs/7447484123?check_suite_focus=true#step:4:95). ~Any idea where `discoverers` might be coming from?~ That derivation seems to be coming from dream2nix. Their README says "Nixpkgs Compatibility, by not enforcing IFD (import from derivation)", so there should be a way around that?

I'm not sure why that's the case, I can build attributes / run `flake show`, `flake check` fine with `--no-allow-import-from-derivation`. My nix version is `2.9.1`.

ncfavier:
So it looks like evaluating e.g. `(builtins.getFlake "github:yusdacra/nix-cargo-integration").packages.x86_64-linux.nci-cli.name` fails with `--read-only` but succeeds with just `--no-allow-import-from-derivation` (and then subsequently succeeds with `--read-only` since the offending path is now in store).

@erikarvstedt any idea why?

I think it's probably better to remove `--read-only` for now.

erikarvstedt:
Just judging by the backtrace of 
```bash
nix eval --show-trace --read-only --expr '(builtins.getFlake "github:yusdacra/nix-cargo-integration/4080369b93d7f35ed90224a4557b11e50230337e").packages.x86_64-linux.nci-cli.name'
```
... I guess that [this line in dream2nix/lib/modules.nix](https://github.com/nix-community/dream2nix/blob/aaa994667f78e2da5cde3f2db1e53926c7905846/src/lib/modules.nix#L146), which evaluates to something like `node2nix/src/subsystems/rust/discoverers/default`, leads to this path getting added to the nix store due to the unqualified inclusion [of the path in a string literal](https://github.com/nix-community/dream2nix/blob/aaa994667f78e2da5cde3f2db1e53926c7905846/src/lib/modules.nix#L152).
 
This is fixed by:
```nix
dir = "${toString ../subsystems}/${subsystem}/${modulesCategory}";`
```
Adding store paths for evaluation is always an antipattern, so it makes sense to keep`--read-only` as a flake sanity check.

ncfavier:
I see, thanks! (maybe adding store paths in this way should be disallowed by `--no-allow-import-from-derivation`?)

ncfavier:
Sadly fixing that line isn't enough since there are *lots* of these antipatterns spread across the dream2nix and nix-cargo-integration codebases and finding where exactly they originate from isn't trivial. I think it's fine to remove `--read-only` for now.

I do wonder though, wouldn't using `toString` result in impurities when dream2nix is not used as a flake? It seems to me like the resulting paths could depend on where dream2nix is located on the filesystem. Isn't avoiding that the whole point of copying things to the store?

erikarvstedt:
> I do wonder though, wouldn't using toString result in impurities when dream2nix is not used as a flake? 

No, not when the paths are only used for importing them during evaluation.

Antipattern:
```nix
import "${./src}/lib.nix"`;
```
Valid use case:
```nix
writers.writeBash "run-app" ''
  exec ${myApp}/bin/app --config ${./src/default-config.json} "$@"
''
```

ncfavier:
Aah I see thanks.

yusdacra:
I'm not sure I understand why this is an antipattern when it is a useful feature we have been leaning on. Doesn't this also mean that using `builtins.path` is also disallowed with `--read-only` since it will write to store at evaluation time? It would also mean that `builtins.toFile` is disallowed when it can be useful in some cases.

Also I'm not sure I understand why `"${toString ../subsystems}/${subsystem}/${modulesCategory}"` works and `../subsystems + "/${subsystem}/${modulesCategory}"` doesn't... Is there some special property of calling `toString` manually? AFAIK it will get coerced to a string with `toString` so shouldn't it behave the same way?

erikarvstedt:
@yusdacra, see [here for details](https://stackoverflow.com/a/43850372/3806468) about string interpolation with paths.

`builtins.path` and `builtins.toFile` are fine to use as long you don't import the resulting paths during evaluation (which you probably don't).

yusdacra:
> @yusdacra, see [here for details](https://stackoverflow.com/a/43850372/3806468) about string interpolation with paths.
> 
> `builtins.path` and `builtins.toFile` are fine to use as long as you don't import the resulting paths during evaluation (which you probably don't).

Thanks for the link, I'll take a look.

About the `builtins`, using `builtins.readFile` / `builtins.pathExists` will also count as "importing" no? We do make use of that in `dream2nix` to read certain file information and to decide certain stuff. But, I was able to fix it I believe, at least enough to make using `flake-info` pass (the revision that uses the `--read-only` flag), https://github.com/nix-community/dream2nix/pull/209.

erikarvstedt:
> using builtins.readFile / builtins.pathExists will also count as "importing"

It will only cause an error with `--read-only` if you generate these paths beforehand while evaluating.

ncfavier:
@yusdacra FWIW your flakes are still failing with `--read-only`. nix-cargo-integration needs a `toString ./cli` on flake.nix:45, and rust-nix-templater is probably still failing because of dream2nix

yusdacra:
> @yusdacra FWIW your flakes are still failing with `--read-only`. nix-cargo-integration needs a `toString ./cli` on flake.nix:45, and rust-nix-templater is probably still failing because of dream2nix

Thanks, I need to fix those as well; just haven't found the time to do so yet.

PR #503, opened 2022-07-22
flake-info: fix compilation warnings

ncfavier:
Removes some dead code, including the unimplemented `--elastic-user`, `--elastic-pw` and `--gc` flags. Builds and tests with no warnings.


ysndr:
`--gc` is used though
You can (could?) run the import on a fresh store which the gc flag would clean up after completion.
I dont know if that is still useful to anyone though and could probably be implemented by just passing extra nix arguments

ysndr:
for the elastic-pw/user one I wonder if it would be better to implement these rather than remove them and use those flags rather than the url encoded authentication

ncfavier:
> `--gc` is used though You can (could?) run the import on a fresh store which the gc flag would clean up after completion. I dont know if that is still useful to anyone though and could probably be implemented by just passing extra nix arguments

It's not used. The `run_gc` function is never called. So I think it's safe to assume it's not useful to anyone :)

> for the elastic-pw/user one I wonder if it would be better to implement these rather than remove them and use those flags rather than the url encoded authentication

Sure, you can do that in a follow-up PR, I'm just commenting them out for now.

ysndr:
I agree in both points

ncfavier:
@ysndr could you approve please?

ysndr (approved):

ncfavier:
Thanks

PR #505, opened 2022-07-24
Revert "flake-info: remove `--read-only` (#504)"

ncfavier:
Reverts NixOS/nixos-search#504, see https://github.com/NixOS/nixos-search/pull/502#issuecomment-1193305553

The issue with some flakes is really a bug with those flakes (needlessly copying paths to the store for evaluation) rather than a bug with Nix.

(sorry for the back-and-forth)


ysndr:
Do i read right that we still want to remove the flag for now since the bug is quite common with flakes using `dream2nix` and friends?

ncfavier:
I don't know. They seem close to being fixed now.

Issue #508, opened 2022-07-27
packages: Offer more installation method suggestions and discourage `nix-env -i`


roberth:
Currently the package details only suggest `nix-env`.

`nix-env` is considered bad practice by many, so we should at least offer better, declarative and/or non-global installation methods in the packages search details.

 - NixOS `environment.systemPackages`
 - home-manager
 - package or nix-shell
    - with `callPackage`
    - with `pkgs`
    - as `nativeBuildInputs`
    - as `buildInputs`

This also solves the problem in docs and conversations of having to explain the general idea of "package installation" over and over again, which involves either a roundtrip to ask what sort of installation they need (might not even be clear) or, in the case of docs, trying to explain multiple methods.
Knowing the attribute path also means that we can show relevant instructions for dealing with nested attributes in `callPackages`.

Related: https://github.com/NixOS/nixos-search/issues/506#issuecomment-1197073259

fricklerhandwerk:
Also `nix-shell -p`.

I fully agree `nix-env` should not be the first or only suggestion. Presenting all these options would need a different UI, e.g. a link to a page listing all the options and linking to a tutorial piece for details.

See [`nix-versions`](https://github.com/lazamar/nix-package-versions) for an [example](https://lazamar.co.uk/nix-versions/?package=hello&version=2.12&fullName=hello-2.12&keyName=hello&revision=bf972dc380f36a3bf83db052380e55f0eaa7dcb6&channel=nixpkgs-unstable#instructions).
It also [shows `nix-env` first](https://github.com/lazamar/nix-package-versions/blob/419b52afc3fabcc3d84d132037d106ad929a5d33/src/App/Server.hs#L276-L308).

roberth:
> Presenting all these options would need a different UI

Presentation of the search results is handled by a single page application written in Elm. It is already set up to display multiple tabs with an installation method in each of them.
I don't see much of a _technical_ reason to build a new kind of page, but it would be nice to have a page that displays only a single package. Linking to the search results can be a bit noisy, both in the url and on screen.

fricklerhandwerk:
Ah, I missed that tab thingy. In that case, great.

PR #514, opened 2022-08-06
encourage NixOS configuration, nix-shell and discourage nix-env

MatthewCroughan:
This would close https://github.com/NixOS/nixos-search/issues/508

# High level changes

1. `nix-env` now has a warning associated with it, which explains that it has side-effects which may require cleanup by the user manually after usage like a traditional package manager.
2. NixOS configuration(s) and nix-shell have now got a tab on each package, which explains how to use `environment.systemPackages` or `nix-shell -p`.

# Internal changes

1. I have replaced `FromNixOS` and `FromNixpkgs` with `ViaNixOS`, `ViaNixShell` and `ViaNixEnv`
2. I have removed the `$` before every instance of `.terminal` in `frontend/src/index.less` because I was not easily able to case split on imperative cmds like `nix-shell -p` and the code for a NixOS configuration. Instead, `$` has been placed manually into each text field where a command is imperative, such as `nix build` for flakes and `nix-shell`.
3. I'm no ELM expert, so have likely made wrong use of the `div` function, and would appreciate someone to help with that

# What it looks like

### NixOS Configuration
image

### nix-shell
image

### nix-env
image

### DIFF
--- a/frontend/src/index.less
+++ b/frontend/src/index.less
@@ -388,8 +388,13 @@ header .navbar.navbar-static-top {
               margin: 2em 0 1em 1em;
               text-align: left;
 
-              // how to install a package
+              // long description of a package
               & > :nth-child(1) {
+                margin-top: 1em;
+              }
+
+              // how to install a package
+              & > :nth-child(2) {
 
                 h4 {
                   font-size: 1.2em;
@@ -415,11 +420,6 @@ header .navbar.navbar-static-top {
                   .terminal();
                 }
 
-              }
-
-              // long description of a package
-              & > :nth-child(2) {
-                margin-top: 1em;
               }
 
               // maintainers and platforms
--- a/frontend/src/index.less
+++ b/frontend/src/index.less
@@ -63,6 +63,10 @@ body {
   cursor: text;
 }
 
+.shell-command:before {
+  content: "$ ";
+}
+
 #content {
   padding-bottom: 4rem;
 }
--- a/frontend/src/index.less
+++ b/frontend/src/index.less
@@ -6,10 +6,6 @@
   background: #333;
   color: #fff;
   margin: 0;
-
-  &:before {
-    content: "$ "
-  }
 }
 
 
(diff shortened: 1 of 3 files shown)


max-privatevoid (approved):

Kranzes (approved):
This change has been approved by the IANEO, the International Anti-nix-env Organization. Consider your work worthy. ✅

frontend/src/index.less
```diff
@@ -6,10 +6,6 @@
   background: #333;
   color: #fff;
   margin: 0;
-
-  &:before {
-    content: "$ "
-  }
```
ncfavier, line 12: The point of having this in a `:before` is that the dollar sign is not copied when you select the whole block

ncfavier (commented):
Looks alright.

Let's maybe move the "long description" above the installation instructions?

Ping @roberth

frontend/src/index.less
roberth: The pre-existing css is the strangest I've ever seen. What's up with all the `nth-child` selectors? Positional as opposed to named css? That's just an awful idea.
So I can't blame Matthew for not restoring this functionality for nix-env. He had to remove the `$` for the declarative installation methods, which are more important anyway.

roberth (approved):

roberth:
An example with a shell file would be nice, but certainly no blocker.

frontend/src/index.less
ncfavier: I know, it's awful. There's no need to stick to that pattern though: adding a `shell-commands` class outside the hierarchy would be fine.

MatthewCroughan: I am not familiar with elm enough to accomplish this. If you could make a PR that does it to my fork, I'll merge it and it'll be done. But I think the scope of this PR will then have been increased more than it needs to be.

ncfavier: No problem! I pushed to your branch directly.

ncfavier (approved):

MatthewCroughan:
@ncfavier Having reviewed your changes, it looks great.

image

fufexan (approved):
Great stuff :)

roberth:
@garbas This looks ready to merge.

PR #516, opened 2022-08-15
Add geomyidae server flake to public flakes

Munksgaard:


Munksgaard:
I'm not exactly sure if this is enough to support sourcehut flakes, let me know if there's anything else that needs to be done. Also, if you want the sourcehut support in a separate PR, let me know.

flake-info/README.md
```diff
@@ -68,7 +68,8 @@ The `<flake>` argument should contain a valid reference to a flake. It accepts a
 
 > use git+<url> to checkout a git repository at <url>
 > use /local/absolute/path or ./relative/path to load a local source
-> use gitlab:<user>/<repo>/github:<user>/<repo> to shortcut gitlab or github repositories
+> use gitlab:<user>/<repo>/github:<user>/<repo>/sourcehut:<user>/<repo> to
+> shortcut gitlab, github or sourcehut repositories
```
ncfavier, line 72: Could you remove the entire quotation block here and add a link to https://nixos.org/manual/nix/unstable/command-ref/new-cli/nix3-flake.html#types instead?

flake-info/README.md
```diff
@@ -97,7 +98,7 @@ $ flake-info nixpkgs nixos-21.05
 
 to perform a bulk import grouping multiple inputs under the same name/index use the group command.
 
-It expects a json file as input that contains references to flakes or nixpkgs. If those resources are on github or gitlab they can be extended with more meta information including pinning the commit hash/ref.
+It expects a json file as input that contains references to flakes or nixpkgs. If those resources are on github, gitlab our sourcehut they can be extended with more meta information including pinning the commit hash/ref.
```
ncfavier, line 101:
```suggestion
It expects a JSON file as input that contains references to flakes or nixpkgs. If those resources are on GitHub, GitLab or SourceHut they can be extended with more meta information including pinning the commit hash/ref.
```

ncfavier (commented):
Few comments but looks fine, we should probably move the SourceHut support to another PR since we do squash merges.

Regarding your flake, the `example` value for `services.geomyidae.base` is triggering a Nix [bug](https://github.com/NixOS/nix/issues/5612). You can work around this for now by setting `example = lib.literalExpression "/var/gopher";`.

One more remark that's probably not a blocker: it's not very easy to find out what geomyidae *is* from your flake. The link to geomyidae in your README is a gopher link so I couldn't click it, googling "geomyidae" turned up a bunch of adorable rodents so I had to add some word like "service" to find out that it was in fact a gopher server. Maybe you could include this information ("the geomyidae gopher server") in your `enable` option's description and/or in your README.

flake-info/README.md
Munksgaard: Ah, but not all types supported by flakes are supported by `nix-search`, are they?

Munksgaard:
I've separated the handling of sourcehut into a separate PR (#517). I'll update the info on my flake as per your suggestion :-)

ncfavier:
Could you rebase this on main please?

Munksgaard:
Done :-)

ncfavier (approved):

PR #517, opened 2022-08-16
Add support for sourcehut links

Munksgaard:
Separated out from #516


ncfavier:
> Ah, but not all types supported by flakes are supported by nix-search, are they?

You're right. I think it would be a big improvement to make flake-info accept any flake reference that Nix accepts, and do its best to support them all.

ncfavier (approved):

PR #528, opened 2022-08-29
Ignore `_module` options for flakes

ncfavier:
Having the documentation for `_module.args` in the nixpkgs options is useful, but there's no need to have a copy of it per flake module.

Accidentally fixes https://github.com/NixOS/nixos-search/issues/534


ysndr (approved):
Yes, _module is not super useful indeed.

PR #529, opened 2022-08-29
Allow elastic feature to be disabled

### LINKED ISSUE #523: Feature request: Make use of rust conditional compilation [feature]
When using flake-info as a dependency, lots of dependencies are included in the lib that the user might not want. Most notably, uploading to elasticsearch is unnecessary for me and has a lot of dependencies.

Rust has the possibility to use conditional compilation using `{features = [...] }` in the `Cargo.toml` (see [doc](https://doc.rust-lang.org/cargo/reference/features.html)).

ysndr:
If using flake-info as a library, the elastic dependency is likely to be undesirable.
This allows to disable the elastic feature and in turn compiles libflox-info without support for elasticsearch

resolves #523

### DIFF
--- a/flake-info/Cargo.toml
+++ b/flake-info/Cargo.toml
@@ -27,7 +27,11 @@ sha2 = "0.9"
 pandoc = "0.8"
 semver = "1.0"
 
-elasticsearch = {git = "https://github.com/elastic/elasticsearch-rs", features = ["rustls-tls"]}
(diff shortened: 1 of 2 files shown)


ncfavier:
Doesn't build with `--no-default-features`

ysndr:
> Doesn't build with --no-default-features
its just the library that is supposed to build without elastic

did you try `cargo build --lib --no-default-features` ?

ncfavier (approved):
Right, that works

PR #530, opened 2022-08-30
Add nix-vscode-marketplace Source

AmeerTaweel:
I tested by running:

```bash
nix run github:nixos/nixos-search#flake-info -- flake github:AmeerTaweel/nix-vscode-marketplace
```

And got no errors.

### DIFF
--- a/flakes/manual.toml
+++ b/flakes/manual.toml
@@ -118,3 +118,8 @@ repo = "haskell.nix"
 type = "sourcehut"
 owner = "~munksgaard"
 repo = "geomyidae-flake"
+
+[[sources]]
+type = "github"
+owner = "AmeerTaweel"
+repo = "nix-vscode-marketplace"


ncfavier:
Your flake doesn't expose any packages or options, so adding this would have no effect. You can see this by adding `--json` to your command.

AmeerTaweel:
The flake exposes packages under a namespace. How can I make Nix Search recognize these?

ysndr:
The `packages` attribute is supposed to be a flat attribute set.
A good first attempt is to make `flake show` work on your flake which will highlight these mismatches. `flake check` might work too to check your flakes structure
One attempt to make a flat package set is to namespace the keys so instead of 

```
{
 a = { 
   b = "somthing";
 };
}
```

you may write:
```
{
  "a/b" = something
}
```

AmeerTaweel:
It seems that `nix-vscode-marketplace` will be merged into `nixpkgs`. So I'm closing this...

PR #536, opened 2022-09-08
Allow `modulesPath` in NixOS modules

ncfavier:
Unblocks https://github.com/NixOS/nixos-search/pull/500#issuecomment-1240709135

### DIFF
--- a/flake-info/src/commands/flake_info.nix
+++ b/flake-info/src/commands/flake_info.nix
@@ -44,7 +44,7 @@ let
     )
   ) apps;
 
-  readOptions = let
+  readNixOSOptions = let
     declarations = module: (
       lib.evalModules {
         modules = (if lib.isList module then module else [ module ]) ++ [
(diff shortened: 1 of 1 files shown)


flake-info/src/commands/flake_info.nix
```diff
@@ -54,6 +54,13 @@ let
             }
           )
         ];
+        specialArgs = {
+          # !!! NixOS-specific. Unfortunately, NixOS modules can rely on the `modulesPath`
+          # argument to import modules from the nixos tree. However, most of the time
+          # this is done to import *profiles* which do not declare any options, so we
+          # can allow it.
+          modulesPath = <nixpkgs/nixos/modules>;
```
ysndr, line 62:
```suggestion
          modulesPath = "${nixpkgs}/nixos/modules";
```

ysndr (approved):

flake-info/src/commands/flake_info.nix
ncfavier: Doesn't work

ncfavier: `nixpkgs.path` does, let's do that.

Issue #537, opened 2022-09-09
Options search doesn't match option prefixes


erikarvstedt:
[extradep](https://search.nixos.org/options?channel=unstable&from=0&size=30&sort=relevance&type=packages&query=extradep) and [extraDep](https://search.nixos.org/options?channel=unstable&from=0&size=30&sort=relevance&type=packages&query=extraDep) should match `system.extraDependencies`, but they don't.
The [full option string](https://search.nixos.org/options?channel=unstable&from=0&size=30&sort=relevance&type=packages&query=extradependencies) matches correctly.

erikarvstedt:
Not fixed.
[extradependencies](https://search.nixos.org/options?query=extradependencies&from=0&size=30&sort=relevance&channel=unstable) matches, but [extradep](https://search.nixos.org/options?channel=unstable&from=0&size=30&sort=relevance&type=packages&query=extradep) doesn't.

ncfavier:
Hopefully when the 7.10 update lands we can use the case_insensitive option and fix this.

PR #538, opened 2022-09-09
Fix wildcard search

ncfavier:
Fixes https://github.com/NixOS/nixos-search/issues/537 maybe.

### DIFF
--- a/frontend/src/Search.elm
+++ b/frontend/src/Search.elm
@@ -1287,7 +1287,7 @@ searchFields query mainField fields =
                   )
                 ]
             )
-            (String.words (String.toLower query))
+            (String.words query)
         )
 
 


ysndr:
We had a wildcard search?

ncfavier:
> We had a wildcard search?

Apparently.

> [wildcard] query does not support [case_insensitive]

Hmm, I guess we're using a version of Elastic older than 7.10.0? (https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-wildcard-query.html)

ncfavier:
Yeah, 7.2.0. Shame.

ncfavier:
Maybe we can just remove the `lowercase` call in the frontend, so that we can at least match things using the proper capitalisation?

ncfavier:
That seems to work. Now at least `extraDep` returns a result.

ysndr:
@garbas is it possible to elevate us onto a newer elastic version?

garbas:
There is `7.10.2` version available, would that work?

ncfavier:
Yes!

garbas:
Sorry I forgot to contact the Bonsai support, I've done it now. I'll keep you posted.

ncfavier:
Maybe we can merge this in the meantime?

garbas:
yeah, lets merge this. works a little better.

I also got news from bonsai (it for some reason went to my spam) that they will bring new cluster with 7.10 up. Then we will have to rework the query a bit.

garbas (approved):

PR #549, opened 2022-10-17
Add dolphin-emu-nix to the flakes index

MatthewCroughan:
Adds an auto updating Flake to the index for dolphin-emu, it is similar to flake-firefox-nightly in what it achieves.


raboof:
It seems dolphin-emu is also in nixpkgs, do you think it makes sense to also have it as a 3rd-party flake?

Issue #486, opened 2022-05-28
Policy for inclusion in the flake search


ncfavier:
Basically a mirror of https://github.com/NixOS/flake-registry/issues/25, triggered by https://github.com/NixOS/nixos-search/pull/482.

I think the policies should be somewhat similar between the two repos (if not tied together by only indexing flakes in the registry), and moreover I don't think nixos-search is currently ready to include every flake in the universe.

ysndr:
I think the issue the registry faces is not one we have in the search.
In the issue you mentioned it was brought about that the registry is a global namespace and therefore has quite a bit more influence.

Only indexing the registry defeats in my opinion the purpose of the search.

A) the search should make projects more discoverable, that includes small projects that are not yet popular enough to be in the registry.
B) As mentioned in the issue, the registry is not really necessary and just a shorthand for a selected set of packages. I don't see why we should bind us to something "unnecessary"


The registry also has to maintain a certain standard of quality and security as packages in there could be considered "officially approved".
Yet, neither us search people nor the maintainers of the registry have the capacity to QA every flake added (and track it over time).
Options to face this are either putting up extensive rules that narrow the amount of flakes that need to be considered/screened or providing a somewhat neutral index.
Considering the goal to improve discoverability I think we should lean more towards the latter although.
Still we should come up with some rules that filter out flakes with no public benefit (such as completely personal configs, or false flag flakes)

Regarding nixos-search, I think indexing more flakes (within limits) is not too much of an issue since most flakes are relatively small.
There are also efforts to include whole organisations into the search.

I agree however that in the current state flakes are quite disorganized and should be addressed in the current redoing.


These are my first thoughts on this
Let's discuss possible filters below or other options below!

ysndr:
Some points of todays discussion on the topic:

Most indexes allow you to publish your work on your own account. **Cargo**, for instance is famously indifferent about the packages listed there.

In contrast, we have a whole PR process for this which may be good to prevent spamming but also makes us discuss the same issues over again to the frustration of flake maintainers.

So what are the reasons we reject/block packages?


- https://github.com/NixOS/nixos-search/pull/530
  - incorrect flake definition 
- https://github.com/NixOS/nixos-search/pull/481: 
  - including may cause more confusion than clarity.
  - too many options?
- https://github.com/NixOS/nixos-search/pull/482:
  - sparked a discussion in the registry repo https://github.com/NixOS/flake-registry/issues/25
    note: we dont have the namespace issue mentioned   
[...]

PR #552, opened 2022-10-28
try to bring back the CI

pinage404:
currently, the CI seems broken https://github.com/NixOS/nixos-search/actions/runs/3307444477/jobs/5484472854

as far as i understand, the error seems to be caused by `github:pinpox/nixos`

### DIFF
--- a/flakes/manual.toml
+++ b/flakes/manual.toml
@@ -45,11 +45,6 @@ repo = "nickel"
 
 [[sources]]
 type = "github"
-owner = "pinpox"
-repo = "nixos"
-
-[[sources]]
-type = "github"
 owner = "Mic92"
 repo = "sops-nix"
 


ncfavier:
CI is not broken. The failure of the "check flake groups" action only indicates that *some* flakes are failing, but the [actual flake import](https://github.com/NixOS/nixos-search/actions/runs/3343470857/jobs/5536720648) still works fine for the flakes that do succeed.

Currently failing flakes are:
- `github:pinpox/nixos` because it relies on module arguments provided by `specialArgs`, which we don't provide here, and uses them in `imports`. Ping @pinpox, a possible workaround would be to remove [this line](https://github.com/pinpox/nixos/blob/619f3ec8f924530104c974bc44934d9a42549ec9/modules/lollypops-common/default.nix#L4) and add `lollypops.nixosModules.lollypops` to [`modules`](https://github.com/pinpox/nixos/blob/619f3ec8f924530104c974bc44934d9a42549ec9/flake.nix#L197) directly.
- `github:srid/emanote` because of IFD (at least), maintainer has been [pinged](https://github.com/NixOS/nixos-search/pull/502#issue-1313017979)
- `sourcehut:~munksgaard/geomyidae-flake`, will be fixed by https://github.com/NixOS/nixos-search/pull/539

IMO defining a proper policy for dealing with failing flakes is something to do after we at least have a [policy for inclusion](https://github.com/NixOS/nixos-search/issues/486), which should also include the technical limitations we have and how to work around some of them. Until then, there's no harm in just leaving them be.

pinage404:
Hi,

Thanks for your quick answer

The red cross and the message `Check failure on line 1 in flakes/manual.toml` misled me, i didn't understood that the CI was not failing

I can close this PR =)

PR #559, opened 2022-11-04
Fix wildcard search (for real this time)

ncfavier:
Fixes https://github.com/NixOS/nixos-search/issues/537

Now that we're on 7.10 we can enable case-insensitive wildcard search.

### DIFF
--- a/frontend/src/Search.elm
+++ b/frontend/src/Search.elm
@@ -1281,6 +1281,7 @@ searchFields query mainField fields =
                         [ ( mainField
                           , Json.Encode.object
                                 [ ( "value", Json.Encode.string ("*" ++ queryWord ++ "*") )
+                                , ( "case_insensitive", Json.Encode.bool True )
                                 ]
                           )
                         ]


ncfavier:
[extradep](https://63645e51644cfa186e9f427a--nixos-search.netlify.app/options?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=extradep), [extraDep](https://63645e51644cfa186e9f427a--nixos-search.netlify.app/options?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=extraDep).

garbas (approved):
:heart: 

The current version of search query tries to imitate the wildcard feature as close as possible. But now with new version of elasticsearch we could use real [wildcard query](https://www.elastic.co/guide/en/elasticsearch/reference/7.10/query-dsl-wildcard-query.html). Which would require some testing and making sure queries are fast enough.

ncfavier:
As far as I understand we [are already](https://github.com/NixOS/nixos-search/blob/ccf5ab4775582f5014edfc785c85a77277a2a20b/frontend/src/Search.elm#L1279-L1289) using a wildcard query, as a fallback. `case_insensitive` is a parameter on the page you linked.

garbas:
Actually you're right. My mind was playing tricks on me :)

Issue #540, opened 2022-09-11
No stable links available for packages


IvarWithoutBones:
I sometimes want to link to a package on the nixos-search page from a different repositories readme or a conversation, but the current URL scheme does not seem very reliable: `https://search.nixos.org/packages?channel=unstable&show=discordchatexporter-cli&type=packages&query=discordchatexporter-cli`.

It would be nice if we could use something like `https://search.nixos.org/packages&show=discordchatexporter-cli`, which could default to the `stable` channel. This makes our links look a lot more reliable, and less needlessly expressive.

ncfavier:
Turns out https://search.nixos.org/packages?query=discordchatexporter-cli already works, does that answer your request?

ncfavier:
I guess we could make `query` default to `show` if not provided...

ncfavier:
https://github.com/NixOS/nixos-search/pull/561

ysndr:
We should then also have a way to *get* the shorter permalink.
currently when you open a search result you still need to trim down the query to make it legible

IvarWithoutBones:
> Turns out https://search.nixos.org/packages?query=discordchatexporter-cli already works, does that answer your request?

It does, thanks! The only thing that I think could be added to this to make the UX a bit smoother is a button to copy a short link, as suggested by @ysndr. Not sure if it'd be preferred to open a seperate issue for that, feel free to close this if you think so.

PR #561, opened 2022-11-04
Make search query default to `show` if given

ncfavier:
This allows for shorter permalinks for displaying a given package, see https://github.com/NixOS/nixos-search/issues/540.

### DIFF
--- a/frontend/src/Search.elm
+++ b/frontend/src/Search.elm
@@ -313,6 +313,9 @@ init args defaultNixOSChannel nixosChannels maybeModel =
       , query =
             args.query
                 |> Maybe.andThen Route.SearchQuery.searchQueryToString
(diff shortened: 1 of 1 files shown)


ncfavier:
Example: https://6364ddfaec3a8876d27709ea--nixos-search.netlify.app/packages?show=tmux

garbas:
I was more thinking of going into the direction of providing a completely new route for packages, eg. `/package/tmux`. And than show package across channels/flakes/... (maybe even with some history).

I'd like to soon work on removing `/packages` and `/options` routes and have only one search interface (what currently flakes is but improved).

garbas (approved):
I do think this change is non invasive and can help in the short term.

Issue #490, opened 2022-06-09
Following links does not reset the `from` parameter in the URL


RasmusRendal:
Steps to reproduce:
1. [Search for `services.xserver` in `unstable`, and navigate to the second page of results](https://search.nixos.org/options?channel=unstable&show=services.xserver.windowManager.default&from=50&size=50&sort=relevance&type=packages&query=services.xserver).
2. In `services.xserver.windowManager.default`, it suggests that you use `services.xserver.displayManager.defaultSession` instead
3. Click the link. If you open in a new tab, the issue does not appear
4. You get the following screen, where you have to click "First" or "Previous" to get to `services.xserver.displayManager.defaultSession`:

![bug](https://user-images.githubusercontent.com/19538856/172779451-0cbd927b-e483-40dc-9deb-b513166725d6.png)

Expected result:

Clicking a link for a new search should take me to the first page of results

ncfavier:
I see two possible solutions:
1. add `&from=0` to all option xref links
2. make `Search.init` not default to the previous model's from and size

@ysndr thoughts?

ysndr:
Yes i think the right behaviour would be to reset `from` to 1 (or zero if we have no results) when the query changes

PR #562, opened 2022-11-04
Do not restore `from` and `size` from the previous model

ncfavier:
Fixes https://github.com/NixOS/nixos-search/issues/490

Ideally the search should be as little stateful as possible: the URL should always reflect the current search parameters.


ncfavier:
Test link: https://6364e1725d21a60308c45881--nixos-search.netlify.app/options?channel=unstable&show=services.xserver.windowManager.default&from=50&size=50&sort=relevance&type=packages&query=services.xserver

ysndr (approved):
Like it, thank you!

garbas (approved):
:100:

Issue #435, opened 2022-02-27
Firefox architecture mismatch between search and nix-env.


spott:
The nix package search has Firefox as being supported by both architectures of darwin (https://search.nixos.org/packages?channel=21.11&show=firefox&from=0&size=50&sort=relevance&type=packages&query=firefox), however when I try and install it, I get a `error: Package ‘Firefox-97.0.1’ in /nix/store/xc5pyr7v39769j1m49zp3xq86s9abfgn-nixpkgs/nixpkgs/pkgs/applications/networking/browsers/firefox/wrapper.nix:351 is not supported on ‘aarch64-darwin’, refusing to evaluate.`

This bug: https://github.com/NixOS/nixpkgs/commit/abbb9ea8485aa375be1c8c68916df7ca8aab5f40 seems to indicate Firefox is *not* supported on darwin, so it looks like the search results are out of date.

PR #580, opened 2022-12-06
Take `meta.badPlatforms` into account

ncfavier:
Some nixpkgs packages like firefox have a [`meta.badPlatforms`](https://github.com/NixOS/nixpkgs/blob/1c9ffcf70786f0966982ce0fc76ec05df2e1dec2/pkgs/stdenv/generic/check-meta.nix#L62) attribute which is supposed to be subtracted from `meta.platforms`.

Fixes https://github.com/NixOS/nixos-search/issues/435

Removes the `System` struct which is not a thing in nixpkgs and not supported by the frontend anyway.

Tested the import for 22.05, 22.11 and unstable.


flake-info/src/data/export.rs
```diff
@@ -224,6 +223,23 @@ impl TryFrom<import::NixpkgsEntry> for Derivation {
...
+                    .collect();
+
+                let bad_platforms: HashSet<String> = package
+                    .meta
+                    .bad_platforms
+                    .map_or(Default::default(), Flatten::flatten)
+                    .into_iter()
+                    .collect();
+
+                let package_platforms: Vec<String> =
```
ysndr, line 240:
```suggestion
                let platforms: Vec<String> =
```

Not a strong opinion, but reassigning platforms (or assigning it in a scope) would avoid using the wrong set later on

flake-info/src/data/import.rs
```diff
@@ -27,7 +26,7 @@ pub enum FlakeEntry {
         attribute_name: String,
         name: String,
         version: String,
-        platforms: Vec<System>,
+        platforms: Vec<String>,
         outputs: Vec<String>,
         default_output: String,
         description: Option<String>,
```
ysndr, line 32: I'd like to see types for the specific attributes (even if its just type aliases to String), but that's for another time

ysndr (approved):

Issue #583, opened 2022-12-19
Exact matches not showing up first


augyg:
In searching for numpy, I noticed that the numpy package is listed way down the page and have seen this for other python packages I've looked for recently like matplotlib. 

![image](https://user-images.githubusercontent.com/36346907/208352229-2a16732f-c736-49f7-97f8-99bf7b3dab74.png)

I believe the numpy package or an exact match of the search query should be the "Best Match" as the filter indicates

ncfavier:
This seems to be a regression caused by https://github.com/NixOS/nixos-search/pull/579, since the version [before that](https://github.com/NixOS/nixos-search/pull/559#issuecomment-1302823951) doesn't have that problem :weary: 

I'll try to get it working with [`word_delimiter`](https://www.elastic.co/guide/en/elasticsearch/reference/7.10/analysis-word-delimiter-graph-tokenfilter.html), otherwise I suppose we should revert that PR for now since it didn't solve any real issue.

PR #584, opened 2022-12-19
Use `path_hierarchy` tokenizer for attr paths

ncfavier:
Fixes https://github.com/NixOS/nixos-search/issues/583 by using the [`path_hierarchy` tokenizer](https://www.elastic.co/guide/en/elasticsearch/reference/7.10/analysis-pathhierarchy-tokenizer.html) to index attribute paths as `.`-delimited strings, both from the left and the right, so that `foo.bar.qux` will produce the tokens `foo foo.bar foo.bar.qux bar.qux qux`.

Also remove `<>` from the characters included in n-grams as those are usually part of XML tags.

### Tests

Packages:
- [numpy](https://63a45194e3fb811803bb362a--nixos-search.netlify.app/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=numpy)
- [python310Packages.numpy](https://63a45194e3fb811803bb362a--nixos-search.netlify.app/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=python310Packages.numpy)
- [codium](https://63a45194e3fb811803bb362a--nixos-search.netlify.app/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=codium)

Options:
- [extradep](https://63a45194e3fb811803bb362a--nixos-search.netlify.app/options?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=extradep)
- [/etc](https://63a45194e3fb811803bb362a--nixos-search.netlify.app/options?channel=unstable&from=0&size=50&sort=alpha_asc&type=packages&query=%2Fetc) now matches `environment.etc` which has `<code>/etc</code>` in its description


flake-info/src/elastic.rs
```diff
@@ -155,14 +159,26 @@ lazy_static! {
...
+                    },
+                    "attr_path": {
+                        "type": "path_hierarchy",
+                        "delimiter": ".",
+                    },
+                    "attr_path_reverse": {
+                        "type": "path_hierarchy",
+                        "delimiter": ".",
+                        "reverse": true,
                     },
```
ysndr, line 176: Didn't we remove those reverse fields recently or is this just allowing us to search attrs in paths from right to left?

ysndr (approved):

flake-info/src/elastic.rs
ncfavier: The reverse fields we had were string reversals computed in Rust, e.g. turning `foo.bar` into `rab.oof` so that a search for `bar` could be reversed to `rab` and then match the prefix generated by `edge_ngram`.

This is a different approach, `reverse` here just means we start splitting the string from the right so that we end up with `bar` and `foo.bar`.

PR #588, opened 2022-12-24
Add buf cli flake to public flakes search

DGollings:


raboof:
this flake has 1.10.0 and hasn't been updated for 3 years, while nixpkgs already carries 1.50.0 - I'll close this but LMK if there's a reason to reopen!

PR #590, opened 2022-12-24
Add cockroachdb binary flake to public flakes search

DGollings:


raboof:
cockroachdb is already in nixpkgs, do you think it makes sense to include the flake in the search?

PR #591, opened 2022-12-24
Add sqlboiler flake to public flakes search

DGollings:


raboof:
This package is already in nixpkgs, do you think it makes sense to also have it as a flake?

DGollings:
It appears I thought so in 2022 :)

PR #593, opened 2023-01-06
Improve accessibility for screenreader users

sempruijs:
People that navigate the web using a screenreader use headings to navigate faster on a website.

With this change, blind users can more easily go to the search results.


ncfavier (approved):
Thanks!

sempruijs:
Thanks! I can navigate it faster now :D

Issue #577, opened 2022-12-05
Improve search for common words such as `make`


roberth:
**Introduce your stance**

Package search should be accessible to beginners.

**Describe the issue**

I've [answered](https://stackoverflow.com/questions/74649163/how-to-install-the-make-command-in-nixos?noredirect=1#comment131787556_74649163) a beginner question, and they couldn't find `gnumake`.

> > Did you try [search.nixos.org/packages](https://search.nixos.org/packages)?
> @ RobertHensing Yes . I've writing make. The only result was ekam

It seems that they were puzzled by the cluttered list of results that comes up for `make`.

I propose three solutions:
1. Return a compact list of package names when the number of results is >1
2. Index the names of binaries, and assign a lot of weight to those keywords. This will probably make `gnumake` float to the top, and make some commands easier to find (e.g. `host`, `mv`, `initdb`)
3. Create a list of hardcoded first results to always include at the top. (might not be as easy to implement and requires more manual work. more of a last resort)

**Page links**

make
https://search.nixos.org/packages?channel=22.11&from=0&size=50&sort=relevance&type=packages&query=make

initdb (postgresql)
https://search.nixos.org/packages?channel=22.11&from=0&size=50&sort=relevance&type=packages&query=initdb

**Additional context**

- [ ] I already created a Pull Request

ncfavier:
I like idea 2. We could use [nix-index](https://github.com/bennofs/nix-index) for this:

- after running `nix-env`, generate the file database with `nix-index` (takes about 5 minutes so this would ~double the nixpkgs import time)
	- note: would be nice to abort the import early in CI if the index for this revision already exists, so that we don't do ~3*10 minutes of useless computation every hour.
- extract a hashmap from attribute names to commands from `nix-locate --at-root --no-group --top-level /bin/ --type x --type s` and merge this information into `NixpkgsEntry.Derivation`
- use this to populate a `package_commands` field in the ElasticSearch backend
- tell the frontend about it

Also we really need to stop exposing the reversed fields to the search, or at least give them a very low weight. I still don't know why we need them.

garbas:
:+1: to use `nix-index`, but does nix-index works with other flakes? can we improve nix-index to also work with other flakes?

> Also we really need to stop exposing the reversed fields to the search, or at least give them a very low weight. I still don't know why we need them.

This was used before the wildcard field was used (2 versions of elasticsearch back).

ncfavier:
[...]

PR #610, opened 2023-02-01
Search programs provided by a package

ncfavier:
Use the `programs.sqlite` database provided with nixpkgs channels to populate a `package_programs` field so that searches for e.g. `make` return `gnumake` with a higher priority.

Fixes https://github.com/NixOS/nixos-search/issues/577


ncfavier:
Looking good: [make](https://63da79e576891a06bdfbf8d9--nixos-search.netlify.app/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=make).

As a nice side-effect this seems to make `bashInteractive` float to the top.

Checked the same things as [last time](https://github.com/NixOS/nixos-search/pull/584#issue-1502741342), no regressions.

roberth:
I [dig](https://63da79e576891a06bdfbf8d9--nixos-search.netlify.app/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=dig) this.

Maybe show the list of programs when expanding the package details?

ncfavier:
> I [dig](https://63da79e576891a06bdfbf8d9--nixos-search.netlify.app/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=dig) this.

Note that dig was already shown first since that's just the package's name.

> Maybe show the list of programs when expanding the package details?

Done.

[kaldi](https://63da963829572f07e9e3499e--nixos-search.netlify.app/packages?channel=unstable&show=kaldi&from=0&size=50&sort=relevance&type=packages&query=kaldi) has a lot of programs.

[roboto](https://63da963829572f07e9e3499e--nixos-search.netlify.app/packages?channel=unstable&show=roboto&from=0&size=50&sort=relevance&type=packages&query=roboto) has none.

roberth:
> > I [dig](https://63da79e576891a06bdfbf8d9--nixos-search.netlify.app/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=dig) this.
> 
> Note that dig was already shown first since that's just the package's name.

I must have been thinking of `command-not-found`.

> > Maybe show the list of programs when expanding the package details?
> 
> Done.

It says "/bin" in the section title which is a little vague. It seems to only list the binaries of the selected output, so perhaps this can be prefixed to the "/bin" title?

It'd be nice for it to list other outputs too, and perhaps the things you need to do to actually get one of the alternative outputs. The latter is non-trivial, as it involves `meta.outputsToInstall` and both an expression- and a CLI use case.


(If you'd like to pursue these things; it's already a nice improvement; maybe a follow-up?)

-------


Example:

### `$bin/bin`

Example usage: `${lib.getBin foo}/cmd`, `environment.systemPackages = [ pkgs.foo ];`

### `$bar/bin`

This output is not installed by default.
Example usage: `${foo.bar}/cmd`, `environment.systemPackages = [ pkgs.foo.bar ];`, `nix profile install foo.bar`

(or maybe it should link to a `foo.bar` packages, as these are effectively distinct packages built by the same derivation) (also maybe it should be `nix-env` for now)

ncfavier:
There's no selected output; the bold one is the *default* output (maybe it should just be listed first...). AFAICT nix-index amalgamates packages from all outputs: see [glib](https://63da963829572f07e9e3499e--nixos-search.netlify.app/packages?channel=unstable&show=glib&from=0&size=50&sort=relevance&type=packages&query=glib) (default `bin`) and [tzdata](https://63da963829572f07e9e3499e--nixos-search.netlify.app/packages?channel=unstable&show=tzdata&from=0&size=50&sort=relevance&type=packages&query=tzdata) (default `out`, programs in `bin`).

We could make `nix-index` add an `output` field to the database; not sure I'll have the motivation to follow up on that here though.

fricklerhandwerk:
This is just lovely, thanks so much!

Issue #542, opened 2022-09-15
Display `outputs` metadata as details


fricklerhandwerk:
search.nixos.org is one of the first things Nix beginners encounter - everyone needs packages.

Presenting them with a level of detail only package developers would usually care about is confusing. It requires quite some knowledge of Nix internals to know what outputs are. Confronting beginners with such things fosters the perception of a steep learning curve, and we want to avoid that.

Here's some random package which happens to have multiple outputs.

![Screenshot 2022-09-15 at 14 49 01](https://user-images.githubusercontent.com/6599296/190407875-ad52aad9-2459-42be-9f2e-d06e0a9eb677.png)

## Proposal

Display the output in the details section.

![Screenshot 2022-09-15 at 14 51 38](https://user-images.githubusercontent.com/6599296/190408595-b9bb0a61-3499-443b-9675-a23ce6bb5f18.png)

Sigmanificient:
This would be trivial do to!

PR #620, opened 2023-02-16
move programs to the top of the metadata display

fricklerhandwerk:
rename to "programs provided", since where they are provided is an implementation detail irrelevant to users of `nix-shell -p`.

I'd actually prefer to list provided programs above the fold, because that is what users will care about much more than e.g. outputs of the derivation, which is also an implementation detail.

Related: https://github.com/NixOS/nixos-search/issues/542

### DIFF
--- a/frontend/src/Page/Packages.elm
+++ b/frontend/src/Page/Packages.elm
@@ -504,7 +504,7 @@ viewResultItem nixosChannels channel showInstallDetails show item =
 
         programs =
             div []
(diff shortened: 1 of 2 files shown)


ncfavier (approved):

ncfavier:
The reason I kept it last is that some programs like [busybox](https://search.nixos.org/packages?channel=unstable&show=busybox&from=0&size=50&sort=relevance&type=packages&query=busybox) have many programs, so this hides the maintainers and platforms a bit. But these are only a handful of outliers I guess.

PR #649, opened 2023-04-24
Improve search results with underscore and dash/hyphen package names

### LINKED ISSUE #644: Improve search quality
## Issue description

An existing package named like "foo-bar" is not returned as a result of searching for "foo_bar", and vice versa.

### Steps to reproduce
Examples:
* [Searching for "zookeeper-mt"](https://search.nixos.org/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=zookeeper-mt) gives no results, even though [there is a "zookeeper_mt" package](https://search.nixos.org/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=zookeeper_mt)

* [Searching for "reorder_python_imports"](https://search.nixos.org/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=reorder_python_imports) gives no results, even though [there are multiple packages matching "reorder-python-imports"](https://search.nixos.org/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=reorder-python-imports)

### Recommendation
* Split terms by punctuation (or at least common word separators like dash/hyphen and underscore) when tokenizing the search string and when indexing packages.

MarcoDaniels:
This improvement creates extra "variants" for the wildcard search allowing us to search (fx) for `nodejs_18` and get the result `nodejs-18_x`. Should resolve #644 that also contains more examples of these punctuation cases.

### DIFF
--- a/frontend/src/Search.elm
+++ b/frontend/src/Search.elm
@@ -1240,6 +1240,18 @@ searchFields query mainField fields =
                     )
                 |> List.concat
 
+        queryWordsWildCard =
+            (String.replace "_" "-" query :: String.replace "-" "_" query :: queryWords)
+                |> List.foldl
+                    (\acc uniques ->
+                        if List.member acc uniques then
+                            uniques
+
+                        else
+                            uniques ++ [ acc ]
+                    )
+                    []
+
         queryWords =
             String.words (String.toLower query)
     in
@@ -1271,7 +1283,7 @@ searchFields query mainField fields =
                   )
                 ]
             )
-            queryWords
+            queryWordsWildCard
         )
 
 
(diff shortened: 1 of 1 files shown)


ncfavier (approved):

ncfavier:
Ideally this should probably be done at the level of the analyser, see https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-index-search-time.html. I don't have time to figure this out right now though.

MarcoDaniels:
> Ideally this should probably be done at the level of the analyser, see https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-index-search-time.html and https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-pattern-replace-charfilter.html. I don't have time to figure this out right now though.

I did gave it a try to modify the analyser, but my skills are better at Elm than Elastic 🙂

PR #651, opened 2023-04-27
Accessibility fixes

### LINKED ISSUE #332: Accessibility Concerns
I just opened #331, and after a moment, I thought I should more broadly ping some of the big accessibility concerns I found using Firefox's accessibility tools, WAVE, and my own knowledge.

1. The NixOS logo that links back to nixos.org has no alt text, and is redundant with the "back to nixos.org" link. Visual users will see those near eachother and be unbothered, but a screen reader should read them roughly as "graphic. link to nixos.org. link to nixos.org." I'd probably recommend merging those parts of the header together as one link, and add the alt text "nixos logo". 
2. The actual search form is missing a label. The button is fine, but the only text labeling the search is the placeholder text, which is not guaranteed to be read by screen readers.
3. The document is missing a language specifier, (ie `<html lang="en">`) which enables devices to read a document in the correct language/voice.
4. The contrast is low between text and background colors for every badge indicating the number of packages with a given license/package set/maintainer/etc on the left-hand side of the page. Light blue links are also low-contrast. I recommend using https://contrast-ratio.com/ to determine colors that would be a better contrast, such as `#015a87` (rgb 1, 90, 135) for blue links. 
5. As I said in the other issue, links could also stand to always be underlined, to be absolutely clear they're links regardless of color.
6. The heading level skips from h1 to h4, which can be a problem for navigation and cause people to assume they missed something at 2 or 3. This might not be a huge problem, though.
7. Firefox's accessibility tools show a lot of errors for keyboard navigation - That said, I'm able to tab through the page and navigate it with only a keyboard. The biggest issue I see there directly is that the ▲▲▲ Hide package details ▲▲▲ link does not visually indicate in any way when it is focused with the keyboard. Also, note that screen readers will read out those triangle characters. I'd either add an aria-label, or perhaps move the triangles to aria-hidden spans - they're purely visual indicators.

I only covered the package search here, not options. Many of the same things will apply to options as well, but I frankly don't have the spoons to go through and check more specifics.

MarcoDaniels:
This pull request fixes some of the major issues related to Accessibility, it also resolves #332.

The fix handles all pages (packages, options, flakes):
- Use `button` instead empty `a` tag for pagination, also fixes contrast-ratio.
- Added logo alt and lang
- Accessibility overrides in `frontend/src/index.scss`:
  - A better fix for this would be to use a newer version of bootstrap (current using version 2)

Used "chrome lighthouse" and "axe devtools" to measure accessibility, screenshots with before and after:

<details>
  <summary>Before</summary>

Screenshot 2023-04-26 at 19 38 08

</details>

<details>
  <summary>After</summary>

Screenshot 2023-04-26 at 20 54 27

</details>

### DIFF
--- a/frontend/src/Main.elm
+++ b/frontend/src/Main.elm
@@ -17,7 +17,15 @@ import Html
         , text
         , ul
         )
-import Html.Attributes exposing (alt, class, classList, href, id, src)
+import Html.Attributes
+    exposing
+        ( alt
+        , class
+        , classList
+        , href
+        , id
+        , src
+        )
 import Json.Decode
 import Page.Flakes exposing (Model(..))
 import Page.Home
--- a/frontend/src/Main.elm
+++ b/frontend/src/Main.elm
@@ -17,14 +17,7 @@ import Html
         , text
         , ul
         )
-import Html.Attributes
-    exposing
-        ( class
-        , classList
-        , href
-        , id
-        , src
-        )
+import Html.Attributes exposing (alt, class, classList, href, id, src)
 import Json.Decode
 import Page.Flakes exposing (Model(..))
 import Page.Home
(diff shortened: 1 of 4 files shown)


ncfavier (approved):
Thank you so much!

PR #653, opened 2023-05-02
Cleanup unused frontend code

MarcoDaniels:
Cleanup code:

- Removed unnecessary jQuery files from `index.html`
- Run [elm-review-unused](https://github.com/jfmengels/node-elm-review#try-it-out) to cleanup unused Elm code.

`elm-test` is not really working as `frontend/tests/Example.elm` is just the example and it seems to be some issue with test-dependencies, will probably tackle that in another PR.


MarcoDaniels:
There's seems to be a little inconsistency on what package manager to use for the frontend. The README recommends using `yarn` to run the project, but the frontend contains a `package-lock.json` generated by `npm`.

Should the references to yarn be replaced with npm?

frontend/src/index.html
```diff
@@ -2,29 +2,24 @@
...
 <head>
-  <meta charset="utf-8" />
-  <meta name="viewport" content="width=device-width, initial-scale=1">
-  <title>NixOS Search</title>
+    <meta charset="utf-8"/>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>NixOS Search</title>
 
-  <script type="text/javascript" src="https://nixos.org/js/jquery.min.js"></script>
+    <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
```
ncfavier, line 9: Can we remove the duplicate `<meta name=viewport`?

ncfavier (approved):

ncfavier:
> Should the references to yarn be replaced with npm?

Not sure, cc @garbas

frontend/src/index.html
MarcoDaniels: Yes, missed that one.

garbas:
> > Should the references to yarn be replaced with npm?
> 
> Not sure, cc @garbas

I think this was there to deal with the transition from old search implementation. I'd say remove but keep listening if somebody complains.

also regarding yarn. i think we were using it with yarn2nix and we can safely switch to npm now since we are using npmlock2nix.

PR #660, opened 2023-06-01
Adds ow-mod-man-flake

ShoosGun:
Add to the manual sources [ShoosGun/ow-mod-man-flake](https://github.com/ShoosGun/ow-mod-man-flake). A flake that allows the installation of the cli and gui versions of the [mod manager](https://github.com/Bwc9876/ow-mod-man) for the game outer wilds.


ShoosGun:
After reading more about the policy in the flake search I now see that this flake might be only usefull to the outer wilds modding community, so I am going to close this pull request.

PR #669, opened 2023-07-03
Bundle `link-manpages.nix`

ncfavier:
It's being [removed](https://github.com/NixOS/nixpkgs/pull/239636#discussion_r1248903109) from nixpkgs as part of the dedocbookification process. Eventually we'll want to use nixos-render-docs to render options.


flake-info/default.nix
```diff
@@ -24,7 +24,7 @@ pkgs.rustPlatform.buildRustPackage rec {
 
   ROOTDIR = builtins.placeholder "out";
   NIXPKGS_PANDOC_FILTERS_PATH = pkgs.path + "/doc/build-aux/pandoc-filters";
```
jtojnar, line 26: This as well.

ncfavier: Thanks, done in https://github.com/NixOS/nixos-search/pull/670

Issue #652, opened 2023-04-27
User Interface improvements


MarcoDaniels:
Hi 👋, after creating some fixes for accessibility in #651 I believe some more improvements can be done to the UI.

Currently the frontend is using a very outdated version of css-bootstrap with some overrides in a `scss` file and since it is using Elm we could "_easily_" replicate a very similar UI with something like [`elm-css`](https://package.elm-lang.org/packages/rtfeldman/elm-css/latest/), or maybe even [`elm-ui`](https://elm-ui.netlify.app/).

Elm-ui provides **a lot** of the building blocks that nixos-search pages are using. I would be glad to create a PR with it soon.

Any thoughts? 💭

Massolari:
I've been using [tailwind](https://tailwindcss.com/) in my Elm applications, and it works great.
The fact that it's opinionated makes it easy to keep the design consistent across the application and it's easily extensible.

PR #689, opened 2023-09-11
Setup elm-css in frontend

MarcoDaniels:
Based on the issue #652 decided to create a PR with the base setup for [elm-css](https://package.elm-lang.org/packages/rtfeldman/elm-css/latest) and _slowly_ replace the `index.scss` file.

Added a couple of examples in `Style.elm` for the loading animation, sidebar buckets and preformatted (pre/code) tags. The plan would be to create a handful of "tokens" to be used across the stylings for colors, spacings, sizes, etc.

If something like this would be approved I would be glad to make more similar PR's and replace all css/bootstrap with elm-css records for better usage.


raboof:
I'll close this for now but feel free to return to this later :heart:

PR #696, opened 2023-10-18
Instead of warning newcomers about pitfalls with nix-env, what if we just not mention it all

zupo:
And some proof-read / local dev fixes.


zupo:
Hey @ncfavier, thanks a ton for maintaining search.nixos.org, it's an absolute gem! I created this PR after hearing a bunch of people complain it's time we really stop showing people to use nix-env, so I went searching where we still do it, and found it here. 

I'm perfectly happy removing any and all proof-read / local dev changes, just let me know which!

raboof:
I'll close this for now but this'd be great to return to

PR #698, opened 2023-10-25
Add ~dblsaiko/nix-extras

2xsaiko:
This adds https://git.sr.ht/~dblsaiko/nix-extras to the index, it's a collection of packages and NixOS modules that I use in my configurations but which are also ready to be used in others'.

### DIFF
--- a/flakes/manual.toml
+++ b/flakes/manual.toml
@@ -114,3 +114,8 @@ repo = "nix-vscode-extensions"
 type = "github"
 owner = "hyprwm"
 repo = "hyprland"
+
+[[sources]]
+type = "sourcehut"
+owner = "~dblsaiko"
+repo = "nix-extras"


raboof (approved):

raboof:
Seems to fail with:
```
              … while evaluating the module argument `pkgs' in "/nix/store/2pnmvlid82hrccm9qz4qxg8isjlw847n-source/common/modules/programs/kakoune":
      
             error: attribute 'pkgs' missing
             at /nix/store/8rv7qar60c9p7hzwfa02syvl4yzrgmsm-source/lib/modules.nix:554:28:
                553|         addErrorContext (context name)
                554|           (args.${name} or config._module.args.${name})
                   |                            ^
                555|       ) (functionArgs f);
```

I get the feeling we've seen this elsewhere as well. Feel free to open an issue if you think it's a problem in flake-info.

2xsaiko:
Yeah, this seems like a bug in flake-info. Looks like the pkgs argument isn't being passed to the module loader for some reason. That module (kakoune) is fine, I use it myself.

2xsaiko:
Looking at it more closely, I believe this is related to usage of pkgs.formats.*, see the issue description.

PR #730, opened 2024-03-04
Add dark theme

zebreus:
This PR adds a dark theme that is activated if the user has [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) set to dark. The light theme is unchanged, the dark theme has some minor UI tweaks, to make it look better with a darker color scheme. The dark theme also does not have color gradients on some buttons, because I do not like them.

You can try nixos-search with darkmode at [search.wirs.ing](https://search.wirs.ing/).

## Screenshots
Some comparison screenshots between dark and light mode

### Package search
darkmode package search
lightmode package search

### Option search
darkmode option search
lightmode option search

### Flakes search
darkmode flakes search
lightmode flakes search

## Remarks

I had to actively fight the ancient bootstrap stylesheet we are using for styling some things. The bootstrap stylesheet is imported from nixos.org but I am pretty sure that it has not been used there for multiple years. As a consequence the CSS here is quite hacky, but we should probably rework the entire design anyways at some point in the future (related issue #286). For now this should be fine.

This PR will resolve #554

PR #733, opened 2024-03-05
Improve dark theme

zebreus:
This PR implements changes to the dark theme (#730) based on user feedback. A preview deployment is available at [search.wirs.ing](https://search.wirs.ing).

## Feedback

I received the following feedback:

- The header bar should have a different color than the main content
- White text for everything has too much contrast 
- The sort dropdown should have a different background color
- The results do not look like links because they are the same color as the text
- Bad contrast of hovered links

This PR addresses all of these points.

## Remarks

While the result is still not optimal, as now all text is grey, there is no easy fix for this. A better solution would be to redesign [search.nixos.org](https://search.nixos.org) to align its styling more closely with [nixos.org](https://nixos.org), but that would require a lot more work.

### DIFF
--- a/frontend/src/index.scss
+++ b/frontend/src/index.scss
@@ -921,9 +921,20 @@ header .navbar.navbar-static-top {
   background: var(--info-label-background);
 }
 
-a {
+// Use the same color for focused and hovered links, as they are underlined anyway
+a,
+a:focus,
+a:hover,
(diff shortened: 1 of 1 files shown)


garbas (approved):

SuperSandro2000:
Now it feels a lot better already. Thanks!

PR #742, opened 2024-03-12
Fix missing assets

zebreus:
This PR will integrate assets that were previously loaded from nixos.org . The new website does not include the assets, so nixos-search looks weird without them.


garbas:
`frontend/assets/img/glyphicons-halflings.png` is missing as well (bootstrap.min.css) needs it.

garbas (approved):

PR #749, opened 2024-03-29
flake-info: Adapt RFC-166 style

dasJ:
RFC-166 is merged and is going to be the default style for nixpkgs.

It makes sense to also have the official documentation look the same so people both get used to the style and can copy-paste sensibly formatted code.

The RFC is very long but since we only deal with primitive values here (no lets, functions and whatnot), the change is very easy and only affects lists and attrsets with one item.

The only other rule I can think of that would be required would be the maximum line length, but I'm honestly too lazy and I don't think this rule would apply very often in option defaults/examples.

I also added 2 more minor fixup commits for attrset names with weird characters in them and I reduced the number of clones in the pretty-printer.

### DIFF
--- a/flake-info/src/data/import.rs
+++ b/flake-info/src/data/import.rs
@@ -145,7 +145,7 @@ impl Serialize for DocValue {
                     md.to_owned()
                 }))
             }
-            DocValue::Value(v) => serializer.serialize_str(&print_value(v.to_owned())),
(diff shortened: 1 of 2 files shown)


dasJ:
cc @garbas maybe?

garbas (approved):

garbas (approved):

garbas:
@dasJ thank you!

PR #751, opened 2024-03-29
add support for new meta.repository field

lolbinarycat:
i did my best to make sure this works, but because of limitations in documentation, compute power, and my own personal knowledge, i wasn't able to test the full thing, nor was i able to implement the feature where the homepage link will be hidden if it is identical to the repository link.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -4,7 +4,7 @@
   nixConfig.extra-substituters = [ "https://nixos-search.cachix.org" ];
   nixConfig.extra-trusted-public-keys = [ "nixos-search.cachix.org-1:1HV3YF8az4fywnH+pAd+CXFEdpTXtv9WpoivPi+H70o=" ];
 
-  inputs.nixpkgs.url = "nixpkgs/nixos-unstable";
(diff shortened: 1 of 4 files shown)


flake.nix
```diff
@@ -4,7 +4,7 @@
   nixConfig.extra-substituters = [ "https://nixos-search.cachix.org" ];
   nixConfig.extra-trusted-public-keys = [ "nixos-search.cachix.org-1:1HV3YF8az4fywnH+pAd+CXFEdpTXtv9WpoivPi+H70o=" ];
 
-  inputs.nixpkgs.url = "nixpkgs/nixos-unstable";
+  inputs.nixpkgs.url = "github:NixOS/nixpkgs?rev=e3652e0735fbec227f342712f180f4f21f0594f2";
```
raboof, line 7: I don't think this should be here

raboof (commented):
one small comment - I don't have a full development setup configured yet so I can't fully test yet either

flake.nix
lolbinarycat: You're probably right, I think this was a hacky workaround to make things run on my local machine.

I've mostly moved on from nix development, so someone else will need to pick up this PR (also I still don't know elm).

lolbinarycat: sure

PR #755, opened 2024-04-14
Add kernel-overlay

andreoss:
- Latest releases of Linux kernel scrapped from kernel.org


garbas (approved):

raboof:
It looks like this flake uses `allow-import-from-derivation` and that `flake-info` doesn't support this.

raboof:
Let's close this for now until that's supported

PR #767, opened 2024-05-01
ci: jq is already part of ci environment

### LINKED ISSUE #766: Results are out of date
It appears the results for SNO are not being updated.

I'm not sure how out of date it actually is, but for example the `lexical` package should have been at 0.6.0 for days but SNO is showing 0.5.2. At the latest I suspect this would have been in unstable on 4/28.

https://search.nixos.org/packages?channel=unstable&show=lexical&from=0&size=50&sort=relevance&type=packages&query=lexical
https://github.com/NixOS/nixpkgs/pull/306624

garbas:
fixes #766


garbas:
I've merged this without any review since the import was broken and want to get this in ASAP.

Issue #734, opened 2024-03-06
Wording when a package doesn't specify meta.platforms is misleading.


iFreilicht:
When opening a package that doesn't specify `meta.platforms`, the message is simply "This package is not available on any platform."

![Screenshot 2024-03-06 at 02 46 38](https://github.com/NixOS/nixos-search/assets/9742635/b791f76c-e02d-4ec4-8e95-7dd07163c278)

This is somewhat misleading, for two reasons. 

1. A user can try to build the package on any platform, even without setting `NIXPKGS_ALLOW_UNSUPPORTED_SYSTEM=1`
2. For `x86_64-linux`, the package will even be cached in hydra, so it's pretty much guaranteed to work.

My optimal solution for this would be to communicate both these caveats, maybe by showing `x86_64-linux` as a supported platform, with an additional disclaimer that the user can try installing the package on any other platform as well, but that it's not guaranteed to work.

Alternatively, the wording could be changed to "This package does not list any supported platforms.", or it might just show `x86_64-linux` as supported, nothing else.

lolbinarycat:
relevant context from Lily Foster in the matrix chatroom:
> it looks like the relevant line to change is <https://github.com/NixOS/nixos-search/blob/a740e60705d3cf137abdc25164e5d437d8297421/frontend/src/Page/Packages.elm#L493>

PR #774, opened 2024-05-16
fix: wording when a package has no platforms list

Aleksanaa:
"This package is not available on any platform" is misleading and confusing, so I changed that.

Related https://github.com/NixOS/nixpkgs/issues/270206
Closes https://github.com/NixOS/nixos-search/issues/734

### DIFF
--- a/frontend/src/Page/Packages.elm
+++ b/frontend/src/Page/Packages.elm
@@ -490,7 +490,7 @@ viewResultItem nixosChannels channel showInstallDetails show item =
                 , div []
                     (List.append [ h4 [] [ text "Platforms" ] ]
                         (if List.isEmpty item.source.platforms then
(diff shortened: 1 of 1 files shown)


wegank (approved):

Aleksanaa:
Wait, I didn't look carefully at how our search works, does it only list x86_64-linux, aarch64-linux, i686-linux, x86_64-darwin and aarch64-darwin? Have we stated this somewhere?

If so, maybe we can change the error message a little better, but I have no idea how.

dasJ:
@garbas this should be ready for merge ;)

wegank:
> If so, maybe we can change the error message a little better, but I have no idea how.

Maybe try deferring `filterPlatforms` to have a non-empty `item.source.platforms` even if it doesn't intersect the five platforms?

PR #785, opened 2024-06-07
flakes: drop problematic sources

wegank:
These public flakes are not indexed on https://search.nixos.org/flakes due to various errors, and can therefore be safely dropped.

### DIFF
--- a/.github/workflows/check-flake-files.yml
+++ b/.github/workflows/check-flake-files.yml
@@ -37,7 +37,7 @@ jobs:
         do
           echo "::group::Group \"$(basename $flake_group .toml)\""
 
-         nix run --accept-flake-config .#flake-info -- group "$flake_group" "$(basename "$flake_group" .toml)" --report
+         nix run --accept-flake-config .#flake-info -- --json group "$flake_group" "$(basename "$flake_group" .toml)" --report
 
           if [[ -f "./report.txt" ]]
           then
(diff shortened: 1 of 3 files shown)


garbas (approved):

astro:
Where can I learn more about the errors so that they can be fixed?

wegank:
I suppose you could create a `flakes/manual.toml` with a single source, and run

```bash
nix run --accept-flake-config .#flake-info -- --json group flakes/manual.toml manual --report
```

astro:
My modules access `config.networking.hostName` which is missing. Are the Flakes' nixosModules no longer evaluated together with the NixOS modules? Advice appreciated.

wegank:
I have no doubt that the module will work with a normal NixOS configuration, but I'm not sure if `config.networking.hostName` has something defined by looking at [flake-info/assets/commands/flake_info.nix](https://github.com/NixOS/nixos-search/blob/main/flake-info/assets/commands/flake_info.nix).

I guess the simplest option is to replace `inherit (config.networking) hostName;` by `hostName = config.networking.hostName or "$HOSTNAME";`, [as was done once](https://github.com/astro/microvm.nix/blob/e3a4dd5b381fb580804105594cc9c71dc45abdb5/nixos-modules/microvm/options.nix#L7) in the repo, and try to make the descriptions independent of `config` or `pkgs` if there are other similar errors.

PR #791, opened 2024-06-10
frontend: Show current nixpkgs commit

dasJ:
About the commit length chosen: This is what git currently does for nixpkgs. The exact value is dynamic: https://stackoverflow.com/questions/18134627/how-much-of-a-git-sha-is-generally-considered-necessary-to-uniquely-identify-a/21015031#21015031

### DIFF
--- a/frontend/src/Search.elm
+++ b/frontend/src/Search.elm
@@ -31,6 +31,7 @@ module Search exposing
     , viewSearchInput
     )
 
+import Array
 import Base64
 import Browser.Dom
 import Browser.Navigation
@@ -39,6 +40,7 @@ import Html
         ( Html
         , a
         , button
+        , code
         , div
         , form
         , h1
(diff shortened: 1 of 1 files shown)


frontend/src/Search.elm
```diff
@@ -1039,33 +1041,49 @@ viewResults nixosChannels model result viewSuccess _ outMsg categoryName =
...
+          (case List.head result.hits.hits of
+            Nothing -> []
+            Just elem ->
+              case Array.get 3 (Array.fromList (String.split "-" elem.index)) of
+                Nothing -> []
+                Just commit ->
+                  [
+                    text "Data from nixpkgs "
+                    , a [ href ("https://github.com/NixOS/nixpkgs/commit/" ++ commit) ]
+                    [ (code [] [ text (String.slice 0 12 commit) ]) ]
```
garbas, line 1082:
```suggestion
                    [ (code [] [ text (String.slice 0 8 commit) ]) ]
```

to match abbreviated commit length.

garbas (approved):

frontend/src/Search.elm
```diff
@@ -1039,33 +1041,49 @@ viewResults nixosChannels model result viewSuccess _ outMsg categoryName =
...
+          ]
+          (case List.head result.hits.hits of
+            Nothing -> []
+            Just elem ->
+              case Array.get 3 (Array.fromList (String.split "-" elem.index)) of
+                Nothing -> []
+                Just commit ->
+                  [
+                    text "Data from nixpkgs "
+                    , a [ href ("https://github.com/NixOS/nixpkgs/commit/" ++ commit) ]
```
Aleksanaa, line 1081: Maybe it would be better to use `https://github.com/NixOS/nixpkgs/tree/<commit>` which shows the files at that point. Opinions?

PR #796, opened 2024-06-11
frontend: Fix magnifying glass position and font

dasJ:
Used to look like this on Chrome:
image

Used to look like this on Firefox:
image

---

Now looks like this on both:
image


Aleksanaa (approved):
Can confirm it works!

wegank (approved):

PR #805, opened 2024-06-26
flakes: split sources

wegank:
This PR splits `flakes/manual.toml` into multiple files under `flakes/manual/`, and modifies `.github/workflows/check-flake-files.yml` so that it only runs on the modified files.

### DIFF
--- a/.github/workflows/check-flake-files.yml
+++ b/.github/workflows/check-flake-files.yml
@@ -20,6 +20,8 @@ jobs:
 
     - name: Checking out the repository
       uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
 
     - name: Setup
       uses: ./.github/actions/common-setup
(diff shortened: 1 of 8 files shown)


dasJ (commented):
Hey, sorry this took me so long to take a look at.

I really like the direction this is going. This does however only address the PR checks and not the actual import. That would be where the actual issues surface when a downstream flake breaks. So I think we should loop over the groups in the import as well, write them to a yaml (maybe? it should at least be concat'able) and bulk-import that into ES.

wegank:
> Hey, sorry this took me so long to take a look at.
> 
> I really like the direction this is going. This does however only address the PR checks and not the actual import. That would be where the actual issues surface when a downstream flake breaks. So I think we should loop over the groups in the import as well, write them to a yaml (maybe? it should at least be concat'able) and bulk-import that into ES.

I've also lost track of the PR, but I think the one-line change in `.github/workflows/import-to-elasticsearch.yml` does exactly that?

wegank:
I'm closing this PR created from a branch because I won't have access to it soon. I'll instead make a PR from a fork once I have time.

PR #897, opened 2025-03-10
chore: update license

bravesasha:


raboof (approved):
IANAL but seems like an improvement to me.

I wonder if we can [drop the years](https://daniel.haxx.se/blog/2023/01/08/copyright-without-years/)

PR #881, opened 2025-01-22
fix: update packaging request template url

donovanglover:
Reported in https://github.com/NixOS/nixpkgs/issues/375410#issuecomment-2605923513

CC: @SigmaSquadron

PR #909, opened 2025-04-09
fix: Remove link to "make a packaging request"

Yeshey:
packaging_request.md template has been removed from nixpkgs github(https://github.com/NixOS/nixpkgs/tree/master/.github/ISSUE_TEMPLATE) after this PR: https://github.com/NixOS/nixpkgs/pull/391926

Following discussion here: https://github.com/NixOS/nixos-search/pull/881, this removes the now broken link to it from search.nixos.org

CC: @raboof

### DIFF
--- a/frontend/src/Search.elm
+++ b/frontend/src/Search.elm
@@ -872,8 +872,6 @@ viewNoResults categoryName =
         [ h2 [] [ text <| "No " ++ categoryName ++ " found!" ]
         , text "You might want to "
         , Html.a [ href "https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#quick-start-to-adding-a-package" ] [ text "add a package" ]
-        , text " or "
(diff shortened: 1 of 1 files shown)


raboof:
I guess we could improve this guidance further, since people have other options (like searching for or building an out-of-nixpkgs flake), but this seems like a good start

Yeshey:
> I guess we could improve this guidance further, since people have other options (like searching for or building an out-of-nixpkgs flake), but this seems like a good start

There are tons of options, from searching flatpaks to snaps to the NUR packages, or directing to the flakes tab, or the unstable channel, tho the message is the same across tabs and channels, and I feel like there are tons of flakes that just aren't added to nixos-search. I think keeping it simple, as is, is a fair approach. Just because I feel anything besides a general "search for out-of-nixpkgs stuff" will inventively leave options out.
Possibly the message in the Flakes Experimental tab could be different tho

PR #916, opened 2025-04-26
flake-info: support importing a single nixpkgs attribute

raboof:
to make testing somewhat faster


Jaculabilis:
If this is intended to make testing faster, it might be a good idea to have this force `--kind package`, since otherwise it will import just the one package attribute... and then thousands of NixOS options, which isn't very fast.

Jaculabilis:
Sure, changes are up at #934

raboof:
superseded by #934

PR #918, opened 2025-04-26
[chore] make it possible to disable pandoc

raboof:
as it's pretty heavy

'draft' because it can't be merged in this form, it'd need to be put behind a feature flag.


raboof:
I guess I won't get to this

PR #922, opened 2025-04-29
Suggest to search on nixpkgs issues when nothing is found

lnk3:
When an user searches for a package not already merged it won't be found.

For example:
> https://search.nixos.org/packages?channel=unstable&query=nibbler

Right now it won't find the package, because it hass just been requested.
But it might be useful to the user to quickly check for the status on nixpkgs.

Therefore I Added a link to `nixpkgs/issues?q=PACKAGENAME` in `viewNoResults`.


raboof (approved):
It'd be great if we had comprehensive "what to do if software is not in nixpkgs" documentation (outside of nixos-search), but for now this seems OK

PR #934, opened 2025-05-15
flake-info: support importing a single nixpkgs attribute via --attr

Jaculabilis:
Alternative to #916, with the additions that:

* When an attribute is provided to `nixpkgs` and `nixpkgs-archive`, `--kind package` is forced.
* The attribute is provided as `--attr <name>` instead of a positional argument. This is to make it easier to add something like `--option` in the future.

### DIFF
--- a/flake-info/src/bin/flake-info.rs
+++ b/flake-info/src/bin/flake-info.rs
@@ -56,6 +56,11 @@ enum Command {
     Nixpkgs {
         #[structopt(help = "Nixpkgs channel to import")]
         channel: String,
+
+        #[structopt(
+            help = "Restrict to importing a single attribute",
+        )]
+        attribute: Option<String>,
     },
 
     #[structopt(about = "Import nixpkgs channel from archive or local git path")]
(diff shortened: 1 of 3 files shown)


flake-info/src/bin/flake-info.rs
```diff
@@ -226,25 +237,29 @@ async fn run_command(
...
         }
-        Command::NixpkgsArchive { source, channel } => {
+        Command::NixpkgsArchive { source, channel, attribute } => {
             let ident = (
                 "nixos".to_string(),
                 channel.to_owned(),
                 "latest".to_string(),
             );
+            // Force --kind package if an attribute was specified
+            let kind = if attribute.is_some() { Kind::Package } else { kind };
```
raboof, line 258: should we show an error message when `--attr` is provided but `--kind` specifies options instead?

raboof (approved):
seems reasonable, one comment/question

flake-info/src/bin/flake-info.rs
Jaculabilis: Good idea.

PR #963, opened 2025-06-28
add flake.nix example to how to install tabs

BridgerB:
See screenshot for a new "flake.nix" tab in the "Host to install <package>?" section:
image

It works with their currently selected channel (example 24.11 selected): 
image

### DIFF
--- a/frontend/src/Page/Packages.elm
+++ b/frontend/src/Page/Packages.elm
@@ -597,6 +597,20 @@ viewResultItem nixosChannels channel showInstallDetails show item =
                                         ]
                                     , li
                                         [ classList
+                                            [ ( "active", showInstallDetails == Search.ViaFlakeNix )
+                                            , ( "pull-right", True )
+                                            ]
+                                        ]
+                                        [ a
+                                            [ href "#"
+                                            , Search.onClickStop <|
+                                                SearchMsg <|
+                                                    Search.ShowInstallDetails Search.ViaFlakeNix
+                                            ]
+                                            [ text "flake.nix" ]
+                                        ]
+                                    , li
+                                        [ classList
                                             [ ( "active", showInstallDetails == Search.ViaNixOS )
                                             , ( "pull-right", True )
                                             ]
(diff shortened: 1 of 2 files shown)


raboof:
I think having flake-based suggestions in there makes sense.

I'm not sure these specific suggestions make sense - how/when would you use it like this? Wouldn't you rather just `nix run nixpkgs#hello`?

BridgerB:
@raboof what specific suggestions do you think would make more sense?

dwt:
@BridgerB I like the approach, but I was actually expecting this to show me how to configure this in a flake based nixos config. 

Thinking about this it would actually be really nice to show all the alternatives for the more modern syntax. That would mean
* `nix profile` instead of `nix-env`
* flake based NixOS configuration (not sure there is actually a meaningful difference? Showing how add the input and perhaps how to instantiate unstable in a flake based config that only has stable would have been really helpful to me when I was starting out though)
* `nix shell` instead of `nix-shell`
* I would also like to see an example of how to use `nix-shell` as a shebang and provide this as a dependency for a script.

I would actually like to have examples for a `shell.nix` and `flake.nix` example that adds this package to a shell. (That is what this pull request adds)

Not quite sure what a great UI would be, but ideally I could specify or at least distinguish which commands / examples I want shown. For my use-case, I like to not use the old commands where possible, especially `nix-env` as it's incompatible with a flake based system setup.

Perhaps the easiest way would be to start out with just adding pull requests to slowly add all the tabs, and then worry about making it easier to use by having a separation between the two worlds?

I would like to help with this, but since I haven't contributed to nixos-search before I am still a bit lost.

dwt:
Thinking about this further, perhaps this would be much better, if the tabs where:

* 'Run Directly'
* 'Try out in shell'
* 'Install imperative' or 'Install from the shell'
* 'Install declarative' or 'Install on NixOS / nix-darwin / home-manager'
* 'Use in shell script'

That would map much more cleanly to the actual use cases (that I have) and would allow grouping the old and new way of doing things inside those tabs much more cleanly.

@maintainers sorry, this should perhaps become it's own discussion / issue at this point?

PR #969, opened 2025-07-06
Format with elm-format

miniBill:
Format all Elm files with `elm-format`


miniBill:
Mistakenly closed, will reopen from the correct branch

PR #971, opened 2025-07-06
Format with elm-format

miniBill:
Format all Elm files with elm-format


Cryolitia (approved):
I don't have the write permission of this repo, just contributed some little functions.

These formats LGTM, maybe we could discuss about a compatible code style and run an auto-lint CI for further PRs.

miniBill:
I mean, there is not much to discuss with regards to code style? Everyone in the ecosystem uses `elm-format`, and it has no configuration, so we should just use that.

I think adding a check in CI could be interesting, but didn't want to shave that yak

Issue #994, opened 2025-08-14
flakes: feat a new "`ref`" field


Malix-Labs:
## References
 
- https://github.com/NixOS/nixos-search/blob/4e41fa526459ba5826f7547228b6b5c43b1154c5/flakes/manual.toml#L13-L15
- https://github.com/NixOS/nixos-search/blob/4e41fa526459ba5826f7547228b6b5c43b1154c5/flakes/manual.toml#L17-L19
- https://github.com/NixOS/nixos-search/blob/4e41fa526459ba5826f7547228b6b5c43b1154c5/flakes/manual.toml#L21-L23
- https://github.com/NixOS/nixos-search/blob/4e41fa526459ba5826f7547228b6b5c43b1154c5/flakes/manual.toml#L25-L27

raboof:
https://nix.dev/manual/nix/2.31/command-ref/new-cli/nix3-flake.html#flake-reference-attributes says `ref` is available on 'several' reference types, but doesn't say whether `github` is one of those. Is it? If so I think this feature would make sense.

Malix-Labs:
> Is it?

Yes for github, I'm using it in my local flake right now

I think all of the providers also does but I'm not sure

Malix-Labs:
**Note**: I previously worked on https://github.com/Malix-Labs/nixos-search/tree/flake-reference-attributes, which would fix this issue, but since #1145 is a better solution, I prefer to not create a PR with code that would get wiped in the _near™_ future

PR #996, opened 2025-08-14
feat(flakes): `ref` field

Malix-Labs:
## Blocking

- #994


raboof:
should we close this for now until you move forward with #994 ?

Malix-Labs:
#1183 would supersede it

This PR is in draft as an example just in case we decide to keep the non-standard TOML flake registry

PR #1003, opened 2025-09-03
frontend: improvements to the default "how to install" layout

dtomvan:
See the message at f8d9c2ccfa3a157d580148ff3c7d4956b37ddaa6 for motivation behind this change.


raboof (approved):
This dropped off my radar, sorry.

I'm not sure we should be promoting `nix profile` either, tbh - but this is an improvement

PR #1006, opened 2025-09-10
highlight mainProgram and show mainProgram even if no programs are known

### LINKED ISSUE #912: Feature request: highlight `meta.mainProgram`, if one is set
Reasoning: this sounds useful, when the program has a lot of executables, but a user wants to just `nix run` it and already know which one will run without looking at the source.

One of design suggestions: make that executable bold
Image

Another way could be adding colour (but not blue, it's not a link/anchor)

elfenermarcell:
This changes the "Programs provided" section.
We have 3 types of packages:
 - has some programs -> show program list as before, highlight mainProgram, if set, with strong and border (resolves #912)
 - no programs provided -> show "This package provides no programs." as before
 - seemingly provides no programs, but has mainProgram set (e.g. unfree packages) -> show "Only the main program of this package is known: "


flake-info/src/data/export.rs
```diff
@@ -73,6 +73,7 @@ pub enum Derivation {
         package_outputs: Vec<String>,
         package_default_output: Option<String>,
         package_programs: Vec<String>,
+        package_mainProgram: Option<String>,
```
raboof, line 76: This top of this file says changes to this file need a bump of the `/VERSION` to make sure a fresh index will be created. Is that not needed for this PR because the change is 'backwards-compatible enough' or an oversight?

raboof (approved):
Cool! One comment on the deployment, but the change LGTM

PR #1014, opened 2025-09-23
feat(flakes): `github:gmodena/nix-flatpak`

Malix-Labs:


Malix-Labs:
It's weird that the check passed for my commit but the checks from the merge commit added afterwards failed

Malix-Labs:
The failing test "automatic-custom-flakes-check" might be caused by #1027

Issue #1033, opened 2025-09-29
fix(cd): flake import


Malix-Labs:
## Problems

Importing flakes is starting to get outscaled

This has started causing CI issues.
For example, https://github.com/NixOS/nixos-search/actions/workflows/import-to-elasticsearch.yml fails repetitively which makes https://search.nixos.org/flakes unable to be updated

### Errors

Disk space seems to be the culprit since errors seems to mostly be

>  `System.IO.IOException: No space left on device`

(except the first failing of the failing streak)

And some other are
> `The job has exceeded the maximum execution time of 6h0m0s`

### Subject

- first failing of the failing streak: [Hourly import to Elasticsearch #30802](https://github.com/NixOS/nixos-search/actions/runs/17957676746)
- last succeeding before the failing streak: [Hourly import to Elasticsearch #30801](https://github.com/NixOS/nixos-search/actions/runs/17956318385)

> https://github.com/NixOS/nixos-search/blob/701750cafad9cef7ca5384324e936df9e365e2d5/.github/workflows/import-to-elasticsearch.yml#L116-L120

That step, and especially for the "manual" matrix group (= "Import manual group" step) is the failing one

## Proposed Solutions

_Ordered by (@Malix-Labs's) preference_

Only one implementation would be sufficient, but a combination could help in case it isn't

1. Enhancing CI algorithm by looping for each flake entry and checking for garbage collecting needs before iterating or by using periodic disk monitoring (which maybe would need to be done in [flake-info](https://github.com/NixOS/nixos-search/tree/main/flake-info?rgh-link-date=2025-10-06T16%3A24%3A42.000Z)
2. Keep a single file but split its evaluation into chunks (preferably of ~ equal hardware requirements, maybe by pre-evaluating each flake hardware requirements before the "Import ${{ matrix.group }} group" / "Import manual group" step)
3. Use [GitHub large runners](https://docs.github.com/en/actions/concepts/runners/larger-runners)
4. Splitting flakes entry into arbitrary files (#1037)

## References

- #1027 (concerns are somewhat related)

Malix-Labs:
@raboof I think #1037 is kind of a leaky workaround because as time goes, more arbitrary split would be necessary 

I guess what would be better is to keep a single file but split it into chunks

raboof:
[...]

Issue #1027, opened 2025-09-23
fix(ci): Flakes Check - Decoupling Patches and Batch


Malix-Labs:
## Subject

https://github.com/NixOS/nixos-search/blob/main/.github/workflows/check-flake-files.yml

## Currently

Currently, [flakes](https://github.com/NixOS/nixos-search/blob/main/flakes/manual.toml) checking is run as a batch operation (checking the whole file) every time through the ["Check Flake Groups" GitHub Action Workflow](https://github.com/NixOS/nixos-search/actions/workflows/check-flake-files.yml) when patches (commits, PRs) are pushed

There is no patch-changes-only (only what has changed in patches) flakes checking

There is no periodic batch flakes checking,
Even though flakes are [imported periodically (currently hourly)](https://github.com/NixOS/nixos-search/actions/workflows/import-to-elasticsearch.yml) with potentially changing sources, without pinning locking flakes between checks (see #1144), nor preventing the individual failing ones from being updated

## Problems

1. We are not periodically batch flakes checking even though flakes are not pinned locked (see #1144) yet updated periodically with potentially changing sources, which is a coverage problem and misleading (batch flakes checking when patch are pushed is not it)
3. Flake checking fails because of flakes that are out of scope of patches
4. Flake checking are wasting compute on out of scope changes
5. Flake checking takes a consequential amount of time to conclude, regardless of patch changes

## Solution

Decouple the Flake Checking CI such that batch (whole file) and patch (only what has changed in patches) flake checking are done separately :

1. **Patch Flake Checking**: only checks the patched lines instead of the whole file
2. **Batch Flake Checking**: periodically (period tbd) checks the whole file

Implementing #1144 could make that easier, since if flakes are pin locked and we cache flake-info results, raised problems would vanish

## Related

- [x] #1144
- [ ] #1182
- [ ] #1183
- [ ] #1033 (architecture concerns are somewhat related)

## References

- https://github.com/tree-sitter/tree-sitter
- https://github.com/tree-sitter-grammars/tree-sitter-toml
- https://github.com/Wilfred/difftastic
- https://github.com/afnanenayet/diffsitter

Malix-Labs:
@raboof the non- pinning locking problems raised in this issue's description has just made me realize I don't understand why [flakes/manual.toml](https://github.com/NixOS/nixos-search/blob/main/flakes/manual.toml) is in toml rather than nix?

We would be able to run `nix flake update` to pin locking them easily, and they would be in nix?

[...]

PR #1037, opened 2025-10-01
Split flake lists into groups

raboof:
Possibly a quick fix for #1033?


.github/workflows/import-to-elasticsearch.yml
```diff
@@ -94,7 +94,9 @@ jobs:
       fail-fast: false
       matrix:
         group:
-          - "manual"
+          - "github-atom"
+          - "github-ntoz"
+          - "other"
```
KiaraGrouwstra, line 99: should these match up with the toml file names, as `manual` did, and `github-atom` does?
because `github-ntoz` seems missing so far, whereas `other` seems not to match `others`?

raboof: > should these match up with the toml file names, as `manual` did, and `github-atom` does?

Correct!

> because `github-ntoz` seems missing so far, whereas `other` seems not to match `others`?

Oops, good eyes, forgot to commit/rename ;) . Now updated.

KiaraGrouwstra (approved):

Malix-Labs:
This also helps #1027

PR #1040, opened 2025-10-07
enhance(ci): step "Create issue if failed" remove dependency

Malix-Labs:


raboof (approved):

Malix-Labs:
I already did :) but:

> that worked on my test repo but could fail on that repo depending perhaps on the organization and repo settings that I have no access to

It's not a big deal don't worry too much but I'd feel bad if I contributed something that doesn't work in this repo, even if it's that small

raboof:
Gotcha, don't worry about it too much, we'll cross that bridge when we get there :)

PR #1060, opened 2025-10-24
feat(flake): `github:viperML/wrapper-manager`

Malix-Labs:
https://github.com/viperML/wrapper-manager


raboof (commented):
LGTM but needs a solution to #1033 first...

Issue #1073, opened 2025-10-31
Missing search results master thread


justDeeevin:
## If you're encountering a missing search result and find yourself here, _do not comment_ with the missing package/option

Something's going on with some packages and options not appearing in search results when they should. I have no idea what the root cause could be, but there are multiple issues right now which all likely stem from some common problem with nix-search, so I thought it would be helpful to create a single issue for it to unify them.

### Relevant issues
- #1070
- #1068
- #1071
- #1067
- #1063

#### Shown on neither 25.05 nor unstable
- `hello`
- `audacity`
- `golangci-lint`
- `uthash`
- `airshipper`

#### Not shown on unstable, shown on 25.05
- `portmod`
- `anvil-editor`
- `kdePackages.qt6gtk2`
- `jetbrains.rust-rover`
- `python3Packages.stack-data`
- `veloren`

#### Only available on unstable, not shown on unstable
- `onetbb`
- `celestegame`

#### Not shown on unstable, shown on 25.05
- `environment`

### What the hell?

Here's some diagnosis progress from #1063:

[from kilianar](https://github.com/NixOS/nixos-search/issues/1063#issuecomment-3469557123):
> I took a closer look at this. The released version of nh (4.2.0) queries nixos-search version 43, while the web UI is already using version 44. When building nh from master, it also uses version 44 and fails to find uthash.
> 
> So it looks like something in nixos-search broke during the transition from version 43 to 44.

[from raboof](https://github.com/NixOS/nixos-search/issues/1063#issuecomment-3472141004):
> So I think the next steps here are to figure out where in the process these results go missing:
> 
> - are they still discovered correctly by flake-info?
> - are they still in the upload to elasticsearch?
> - are they in the elasticsearch index?
> - are they in the query results?
> 
> Anyone?

[from kilianar](https://github.com/NixOS/nixos-search/issues/1063#issuecomment-3472506043):
> This result isn’t included in the query results. I reproduced the request the frontend (and nh search) makes with a small Python script. Searching for `rust-rover`
> - using index `latest-43-nixos-unstable`: the expected result is returned.
> - using index `latest-44-nixos-unstable`: I got zero hits:
> …

eclairevoyant:
This is essentially #1069/#1072 it seems.

raboof:
[...]

PR #1076, opened 2025-11-02
Detect and log Elasticsearch bulk request partial failures

dyegoaurelio:
Elasticsearch bulk API returns HTTP 200 even when individual items fail. The previous code only checked for HTTP-level exceptions, which could cause silent failures that may lead to missing packages in the search index.

Related: #1073

### DIFF
--- a/flake-info/src/elastic.rs
+++ b/flake-info/src/elastic.rs
@@ -272,12 +272,40 @@ impl Elasticsearch {
                 .await
                 .map_err(ElasticsearchError::PushError)?;
 
-            dbg!(response)
-                .exception()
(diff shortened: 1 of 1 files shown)


raboof (approved):
This is an excellent improvement, thanks!

I think we should not just log warnings, but actually make sure the command fails in this case, so that it's obvious that something is wrong.

dyegoaurelio:
@raboof yeah, I think so too.

I didn't want to do it in this PR because I'm not sure if this is the cause for the current issues, so my goal here is just to get more debug information to help troubleshoot it

PR #1088, opened 2025-11-10
run flake import with garbage collection between every group member e…

raboof:
…valuation

we might want to change this to 'gc every X flakes', but that can be a future improvement


dyegoaurelio:
I think it should be 'gc every X flakes'.

Running the job with gc after every flake nearly doubled the execution time

raboof:
> I think it should be 'gc every X flakes'.
> 
> Running the job with gc after every flake nearly doubled the execution time

yeah I suppose so... though doubling the execution time is still better than failing entirely :)

Issue #1094, opened 2025-11-11
fix version check to work with nix nightlies


raboof:
Nix nightlies have a version string of the from `2.32pre20250919_4384ec24` which trips up `flake-info/src/commands/nix_check_version.rs`

PR #1112, opened 2025-12-07
pass `--with-gc` to PR validation workflow as well

raboof:


raboof:
hmm no more out of disk space but other errors. my further testing is blocked by #1094 at the moment though :)

PR #1143, opened 2026-02-14
feat(flake-info): fallbacks

Malix-Labs:
This adds fallbacks to flake-info such that it will try to build only packages / apps / options if one fails

Note that "if the other fail" means that the added flakes are defectuous (e.g. many modules read `config.*` at evaluation time (e.g. `nix-bitcoin`) which is invalid when you don’t have a full NixOS configuration ; they should instead use `mkIf` / `mkOption` or read config in places that only evaluate when options are realized)

I personally don't know if this should be merged or if we instead should rightfully error when flakes are defectuous (and we just comment out the flakes that fails until they are fixed instead), RFC

### DIFF
--- a/flake-info/src/lib.rs
+++ b/flake-info/src/lib.rs
@@ -12,7 +12,7 @@ pub mod data;
 pub mod elastic;
 
 pub use commands::get_flake_info;
-use log::trace;
+use log::{trace, warn};
 
 lazy_static! {
     static ref DATADIR: PathBuf =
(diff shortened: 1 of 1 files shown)


raboof:
> I personally don't know if this should be merged or if we instead should rightfully error when flakes are defectuous (and we just comment out the flakes that fails until they are fixed instead), RFC

The risk I see with this approach is that a flake that used to work might 'break' and switch to the fallback without us noticing. I think we would like to notice and be able to fix the issue, instead of having it disappear silently.

Still, for flakes known to have this issue I think it would be nice to be able to ingest them like this on a per-flake opt-in basis.

raboof:
> Then maybe not erroring, but throwing a warning and creating a GitHub issue linking to it?

Sounds reasonable to me. I'll close this PR, that other approach could be introduced in a new PR.

PR #1155, opened 2026-02-23
fix(nixvim): wrong syntax, change owner

repparw:
source singular instead of sources

also updated owner to nix-community


raboof:
> source singular instead of sources

interesting, we should probably error on that :)

PR #1160, opened 2026-03-07
fix(packages): directly filter `package_platforms` when querying

### LINKED ISSUE #570: [question] Why all node modules don't have x86-64?
Only have i686-linux, aarch64-linux, aarch64-darwin. Like <https://search.nixos.org/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=prettier>

epixtm:
While querying packages, Elasticsearch terms aggregations are hard-capped at `size=20` returned buckets for responses. As bucket fields not in `platforms` can pollute the query response when their `doc_count` overshadows that of the allowlists' counts, this may cause platforms which would otherwise be counted when filtering the `package_platforms` aggregation with `filterPlatformsBucket` to be missed.

A possible hack is to bump `size` > 20. However, this is both stressful on the Elasticsearch instance for large values and dependant on the replacement magic being strictly less than the total possible number of platforms.

This patch introduces additional configurability for terms aggregations, allowing queries to optionally enforce an include list to filter at query time. Additionally, the `size` cap for returned buckets can now be configured per aggregation, if so desired.

Fixes #570.

### DIFF
--- a/frontend/src/Page/Packages.elm
+++ b/frontend/src/Page/Packages.elm
@@ -307,14 +307,9 @@ viewBuckets bucketsAsString result =
             selectedBucket.teams
         |> viewBucket
             "Platforms"
-            (result.aggregations.package_platforms.buckets |> sortBuckets |> filterPlatformsBucket)
+            (result.aggregations.package_platforms.buckets |> sortBuckets)
             (createBucketsMsg .platforms (\s v -> { s | platforms = v }))
             selectedBucket.platforms
-
-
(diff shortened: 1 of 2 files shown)


epixtm:
@garbas Any chance of this being reviewed (or who I should probe for this to make progress)? I was planning on asking in Matrix but haven't had the chance.

raboof (approved):

raboof:
sorry for missing this

PR #1177, opened 2026-04-08
feat(flakes): `github:serokell/*`

Malix-Labs:


Malix-Labs:
A lot of them are failing

PR #1186, opened 2026-04-10
allow searching options of modular services

### LINKED ISSUE #1167: allow searching options of modular services
the recently added [modular services](https://nixos.org/manual/nixos/stable/#modular-services) has options that are [already reflected](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/misc/documentation/modular-services.nix) in the [manual's `options.html`](https://nixos.org/manual/nixos/stable/options#opt-_imports_=___pkgs.ghostunnel.services.default___).
in nixos-search however, such options currently [cannot be found](https://search.nixos.org/options?channel=unstable&query=ghostunnel) yet.

as this feature is likely to find support among more services in the future, one potential enhancement here would be to ensure such options will show up in the search as well.

edit: it has been argued these [should be handled from a separate page](https://github.com/NixOS/nixpkgs/issues/432550#issuecomment-3517636577)

KiaraGrouwstra:
closes #1167.

note that the introspection-based approach here may also be of interest as inspiration for the nixos manual, to deprecate the hacky `documentation.nixos.extraModules` (that's currently cause for a manual extra step when adding new modular services, while authors in practice may and do forget).

disclaimer i used a coding agent in the creation of this patch.


version.nix
```diff
@@ -2,11 +2,11 @@
...
   */
-  import = "44";
+  import = "45";
 
   /**
     Frontend index version used by the UI when querying Elasticsearch
     Keep this at the old version while 'import' populates a new index, then update to switch traffic
   */
-  frontend = "44";
+  frontend = "45";
```
raboof: please read the comment :)

raboof:
we need to split this PR into two: one that changes the backend and changes only the backend 'version', and (once the backend cache has been repopulated) on that updates the frontend and points it to the now-populated new backend.

raboof:
but while prototyping having it in one branch for now might make sense

KiaraGrouwstra:
superseded by #1189.

PR #1187, opened 2026-04-10
chore: update elastic search version

KiaraGrouwstra:


raboof:
This went through #1189 and #1192

PR #1188, opened 2026-04-10
clickable option groups

### LINKED ISSUE #798: navigate option groups
one feature for which i find myself sometimes still using the (outdated?) mynixos.com is the ability to navigate to option groups, e.g. for [`programs.firefox`](https://mynixos.com/nixpkgs/options/programs.firefox).
at https://search.nixos.org/options, one might explicitly search for such a path, but cannot so much click-navigate into such a group.
it might be nice to have something like that for nixos-search as well.

KiaraGrouwstra:
in the option search, has the `Name` attribute form a breadcrumb of submodule groups that may be clicked to navigate to the option group in question, i.e. filter option search to options part of the submodule.

closes #798.

disclaimer i used a coding agent in the creation of this patch.


raboof (approved):
very cool!

PR #1189, opened 2026-04-10
feat: support modular services (backend)

KiaraGrouwstra:
Adds modular-service discovery to flake-info: each package's nixpkgs service modules are introspected into standalone option docs tagged with `type=service` plus service_package/service_module metadata, and package docs gain a package_modular_services list of provided modules.

Bump backend index schema to 45 so ingest writes the new docs to fresh indices. The frontend schema version stays at 44 and continues to read the existing latest-44-* indices until a follow-up PR switches it over.

split out from #1186.

first part of #1167.

### DIFF
--- a/flake-info/src/data/export.rs
+++ b/flake-info/src/data/export.rs
@@ -85,6 +85,7 @@ pub enum Derivation {
         package_system: String,
         package_homepage: Vec<String>,
         package_position: Option<String>,
+        package_modular_services: Vec<String>,
     },
     #[serde(rename = "app")]
     App {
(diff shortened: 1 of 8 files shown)


flake-info/src/data/export.rs
```diff
@@ -85,6 +85,8 @@ pub enum Derivation {
         package_system: String,
         package_homepage: Vec<String>,
         package_position: Option<String>,
+        #[serde(skip_serializing_if = "Vec::is_empty")]
```
raboof: do we need this? we don't do it for other fields?

flake-info/src/data/import.rs
```diff
@@ -238,6 +253,7 @@ arg_enum! {
         App,
         Package,
         Option,
+        Service,
```
raboof: Is this all services or the 'modular' services?

raboof (approved):
I haven't reviewed everything in depth but trust your judgement. Two small comments, then I think this it OK to merge.

flake-info/src/data/import.rs
KiaraGrouwstra: modular ones, renamed now to clarify

KiaraGrouwstra:
n.b. [performance implications](https://github.com/NixOS/nixpkgs/pull/509117#pullrequestreview-4094386807) of the used introspection approach

mweinelt:
This broke the unstable options listing. PTAL

PR #1192, opened 2026-04-11
feat(frontend): show modular service options

### LINKED ISSUE #1167: allow searching options of modular services

KiaraGrouwstra:
Add modular service options to the existing Options page. Use checkboxes to filter between the two option categories, which introduces query parameters `?include_{nixos,modular_service}_options}=0|1` (default 1).

Unifies handling of breadcrumbs (#1188) across these categories .

split out from (and supersedes) #1186.
follow-up of #1189.
closes #1167.

### DIFF
--- a/frontend/src/Main.elm
+++ b/frontend/src/Main.elm
@@ -39,6 +39,7 @@ import Search
         , decodeNixOSChannels
         , defaultFlakeId
         )
+import Set
 import Shortcut
 import Task
 import Url
@@ -176,7 +177,25 @@ attemptQuery (( model, _ ) as pair) =
 
         Options searchModel ->
             if Search.shouldLoad searchModel then
-                submitQuery OptionsMsg Page.Options.makeRequest { searchModel | searchType = OptionSearch }
+                Tuple.mapSecond
+                    (\cmd ->
+                        Cmd.batch
+                            [ cmd
+                            , Cmd.map OptionsMsg <|
+                                Page.Options.makeRequest
+                                    model.elasticsearch
+                                    model.nixosChannels
+                                    OptionSearch
+                                    searchModel.channel
+                                    searchModel.query
+                                    searchModel.from
+                                    searchModel.size
+                                    searchModel.buckets
+                                    searchModel.sort
+                                    searchModel.excludedOptionSources
+                            ]
+                    )
+                    pair
 
             else
                 noEffects pair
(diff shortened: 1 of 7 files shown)


raboof:
Cool! So for example a ghostunnel option now looks like:

out

TBH that doesn't really show me how to actually use this modular service - would it be possible to make that clearer somehow?

KiaraGrouwstra:
what would you consider as a good approach for this?

as per the [manual](https://nixos.org/manual/nixos/stable/#modular-services), such an example could look like:

```nix
{
  system.services.my-service-instance = {
    imports = [ pkgs.some-application.services.some-service-module ];
    foo.settings = {
      # ...
    };
  };
}
```


now, the missing piece there compared to the option path listed in the option search entries seems the import.
the import however also depends on the package used.

that maybe raises the question where to list that, and whether that might mean per-option boilerplate sent over the wire.

options may have e.g. `example` values, which could be written out to the point of a full-fledged usage example. however, usage needs may not be restricted to options that have set such fields.

then, some services are offered by multiple packages, raising questions on how to handle this as well. e.g., should we have say tabs allow swapping out the package name? just link to the general docs?

~~also i just realized the `.default` there now is wrong~~

KiaraGrouwstra:
maybe i should list them as `<imports = [ pkgs.ghostunnel.services.default ]>.ghostunnel.package` again like the manual does, but not sure that satisfies the clarity you're looking for.

raboof:
> what would you consider as a good approach for this?
> 
> as per the [manual](https://nixos.org/manual/nixos/stable/#modular-services), such an example could look like:
> 
> ```nix
> {
>   system.services.my-service-instance = {
>     imports = [ pkgs.some-application.services.some-service-module ];
>     foo.settings = {
>       # ...
>     };
>   };
> }
> ```

This looks great to me!
 
> now, the missing piece there compared to the option path listed in the option search entries seems the import. the import however also depends on the package used.
> 
> that maybe raises the question where to list that, and whether that might mean per-option boilerplate sent over the wire.
> 
> options may have e.g. `example` values, which could be written out to the point of a full-fledged usage example. however, usage needs may not be restricted to options that have set such fields.
> 
> then, some services are offered by multiple packages, raising questions on how to handle this as well. e.g., should we have say tabs allow swapping out the package name? just link to the general docs?

Hmm, introducing another layer of tabs might get confusing as well. Might be fine to just link to general docs in that case.

> maybe i should list them as `<imports = [ pkgs.ghostunnel.services.default ]>.ghostunnel.package` again like the manual does, but not sure that satisfies the clarity you're looking for.

As a user I would not know what to do with that - so *maybe* if you can come up with good documentation/links on how to interpret that, but the above looks a lot more actionable.

KiaraGrouwstra:
image

raboof (approved):
This looks pretty great.

I think the main thing missing would be that it's unclear where to stick the 'imports'. A slightly more fleshed-out example fragment that shows the user where to put that, i.e.

```nix
  system.services.<name> = {
    imports = [ pkgs.php.services.default ];
    php-fpm = {
      settings = ...
    };
  };
```

would be a good extension I think - but if that's difficult somehow then I'm OK to merge as-is as well.

KiaraGrouwstra:
image

raboof:
great!

PR #1196, opened 2026-04-12
fix: revert frontend schema version to 44

KiaraGrouwstra:
PR 1189 bumped both `import` and `frontend` versions to 45, but the frontend should have stayed at 44 until the v45 indices were fully populated. This broke the unstable options listing.

Fixes https://github.com/NixOS/nixos-search/pull/1189#issuecomment-4232732541.

### DIFF
--- a/version.nix
+++ b/version.nix
@@ -8,5 +8,5 @@
     Frontend index version used by the UI when querying Elasticsearch
     Keep this at the old version while 'import' populates a new index, then update to switch traffic
   */
-  frontend = "45";
+  frontend = "44";
 }


raboof (approved):

raboof:
This will keep the new frontend code but point it to the old backend index - are you confident that will work?

raboof:
Let's try it I suppose.

In the mean time I'm looking into completing the population of 45, but that'll take some hours.

KiaraGrouwstra:
cool - when that's worked i guess we should be able to reapply the version update over here again as well.

raboof:
> PR 1189 bumped both `import` and `frontend` versions to 45, but the frontend should have stayed at 44 until the v45 indices were fully populated. This broke the unstable options listing.
> 
> Fixes [#1189 (comment)](https://github.com/NixOS/nixos-search/pull/1189#issuecomment-4232732541).

(for reference it was #1192 that bumpted the frontend to 45)

Issue #1197, opened 2026-04-12
"Bad Status Server returned 404" when searching in https://search.nixos.org/(options/packages) with unstable channel


kagurazaka-ayano:
Reproducible url: https://search.nixos.org/packages?channel=unstable&include_modular_service_options=1&include_nixos_options=1&query=users

Screenshot:

<img width="1179" height="225" alt="Image" src="https://github.com/user-attachments/assets/7ff2a87a-13e6-4c03-8644-d71677bd123a" />

With 25.11 channel there are plenty:

<img width="1219" height="536" alt="Image" src="https://github.com/user-attachments/assets/25bc4771-c21c-43b4-8b74-ddeb3d047359" />

I observed there is a 404 post request:

<img width="1270" height="880" alt="Image" src="https://github.com/user-attachments/assets/578502a6-ca63-4ae0-a6f6-7f9abba65625" />

Request payload:

```json
{"from":0,"size":50,"sort":[{"_score":"desc","option_name":"asc"}],"aggs":{"all":{"global":{},"aggregations":{}}},"query":{"bool":{"filter":[{"terms":{"type":["option","service"],"_name":"filter_option_service"}}],"must_not":[],"must":[{"dis_max":{"tie_breaker":0.7,"queries":[{"multi_match":{"type":"cross_fields","query":"users","analyzer":"whitespace","auto_generate_synonyms_phrase_query":false,"operator":"and","_name":"multi_match_users","fields":["option_name^6","option_name.*^3.5999999999999996","option_description^1","option_description.*^0.6","flake_name^0.5","flake_name.*^0.3","service_package^3","service_package.*^1.7999999999999998","service_packages^3","service_packages.*^1.7999999999999998"]}},{"wildcard":{"option_name":{"value":"*users*","case_insensitive":true}}}]}}]}}}
```

TheRealGramdalf:
Also happening for me with https://search.nixos.org/packages?channel=unstable&include_modular_service_options=1&include_nixos_options=1&query=kanidm
```json
[...]

PR #1199, opened 2026-04-13
elasticsearch bulk upload in smaller chunks

raboof:
since #1189 it seems the unstable crossed some threshold and is causing gateway timeouts while performing bulk upload calls. If that problem persists then let's see if splitting the upload in smaller chunks helps.

Seen twice now - I just started a third attempt, let's see if that times out as well, if so try with this change.

Towards #1197


raboof:
Hmm it succeeded the third time... but maybe we should still make the chunks smaller as it seems we're just on the edge.

tomberek (approved):

PR #1201, opened 2026-04-13
feat(flake-info): add backend support for home-manager options

KiaraGrouwstra:
Add `home-manager-option` as a new indexed document type. The Nix extraction script detects flakes whose store path contains `modules/modules.nix` (home-manager's layout) and evaluates their module system to produce option declarations.

Changes:

- Factor `cleanUpOption`/`substFunction`/`mkDeclaration` out of `readNixOSOptions` for reuse by `readHomeManagerOptions`
- Import home-manager's `lib/stdlib-extended.nix` so `lib.hm.*` helpers are available during module evaluation
- Use `lib.mkForce false` for `_module.check` to override home-manager's own `true` default without conflict
- Add `HomeManagerOption` variant to `FlakeEntry`, `Kind`, and `Derivation` enums with serde tag `home-manager-option`
- Include `home-manager-options` in the `all` extraction output so the Group command picks them up automatically
- Bump import schema version to 46

### DIFF
--- a/flake-info/assets/commands/test/expected-outputs/agenix.json
+++ b/flake-info/assets/commands/test/expected-outputs/agenix.json
@@ -83,7 +83,10 @@
   },
   {
     "declarations": ["modules/age.nix"],
-    "default": { "_type": "literalExpression", "text": "{ }" },
+    "default": {
+      "_type": "literalExpression",
+      "text": "{ }"
+    },
     "description": "Attrset of secrets.\n",
     "entry_type": "option",
     "flake": [
@@ -133,7 +136,10 @@
   },
   {
     "declarations": ["modules/age.nix"],
-    "default": { "_type": "literalExpression", "text": "\"0400\"" },
+    "default": {
+      "_type": "literalExpression",
+      "text": "\"0400\""
+    },
     "description": "Permissions mode of the decrypted secret in a format understood by chmod.\n",
     "entry_type": "option",
     "flake": [
@@ -168,7 +174,10 @@
   },
   {
     "declarations": ["modules/age.nix"],
-    "default": { "_type": "literalExpression", "text": "\"0\"" },
+    "default": {
+      "_type": "literalExpression",
+      "text": "\"0\""
+    },
     "description": "User of the decrypted secret.\n",
     "entry_type": "option",
     "flake": [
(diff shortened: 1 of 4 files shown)


flake-info/assets/commands/test/expected-outputs/agenix.json
```diff
@@ -234,221 +252,17 @@
...
-    "default": { "_type": "literalExpression", "text": "\"/run/agenix.d\"" },
-    "description": "Where secrets are created before they are symlinked to {option}`age.secretsDir`\n",
-    "entry_type": "option",
-    "flake": [
-      "github:ryantm/agenix/fcdea223397448d35d9b31f798479227e80183f6",
-      "age"
-    ],
-    "internal": false,
-    "loc": ["age", "secretsMountPoint"],
-    "name": "age.secretsMountPoint",
```
raboof, line 248: so these are duplicates that are no longer included? seems like an improvement.

flake-info/src/data/import.rs
```diff
@@ -47,6 +47,9 @@ pub enum FlakeEntry {
     },
     /// an option defined in a module of a flake
     Option(NixOption),
+    /// a home-manager option extracted from a flake's module system
+    #[serde(rename = "home-manager-option")]
```
raboof, line 51: (right, so the others are converted with the `rename_all = "lowercase"` but this one needs an explicit rename to make it kebab-case)

flake-info/src/data/import.rs
```diff
@@ -264,6 +268,7 @@ impl AsRef<str> for Kind {
             Kind::App => "apps",
             Kind::Package => "packages",
             Kind::Option => "options",
+            Kind::HomeManagerOption => "home-manager-options",
```
raboof, line 271: we could ask ourselves if/how this scales, but I think it's OK, since we'd maybe want to treat nix-darwin as another special case and then beyond that there should be no further "special kinds" of options, right?

raboof (approved):

flake-info/src/data/import.rs
KiaraGrouwstra: perhaps more might pop up but i wouldn't expect many, yeah

KiaraGrouwstra: on this, a different, apparently more scalable approach seems taken by [NuschtOS search](https://search.nüschtos.de/) ([source](https://github.com/NuschtOS/search))

PR #1208, opened 2026-04-16
fix: better querying by attribute

raboof:
So `flake-info --json nixpkgs unstable --attr liberation_ttf` also works


tomberek (approved):
works as expected

PR #1216, opened 2026-04-20
fix(frontend): point "Declared in" for flake options at the flake repo

### LINKED ISSUE #1215: "Declared in" link for 3rd-party flake options points to nixpkgs instead of the flake's repository
On the `3rd-party Flakes` search, when an option result is expanded, the
`Declared in` field renders a link to
`https://github.com/NixOS/nixpkgs/blob/<branch>/<path>` even though the option
comes from a flake that is not nixpkgs.

For example in

Image

`modules/age.nix` points to `https://github.com/NixOS/nixpkgs/blob/nixos-unstable/modules/age.nix`.

aciceri:
Closes #1215 

Some github flakes encode `option_source` as a free-form sentence (e.g. for `nix-community/autofirma-nix` is `flake.nix, via
 ▏ option flake.nixosModules.autofirma`).
For values that don't look like a file path we render plain text rather than an url.

### DIFF
--- a/frontend/src/Page/Options.elm
+++ b/frontend/src/Page/Options.elm
@@ -85,6 +85,7 @@ type alias ResultItemSource =
     , flakeName : Maybe String
     , flakeDescription : Maybe String
     , flakeUrl : Maybe String
+    , flakeRevision : Maybe String
 
     -- modular service metadata (populated only for `service` docs)
     , servicePackage : Maybe String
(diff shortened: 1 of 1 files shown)


frontend/src/Page/Options.elm
```diff
@@ -529,6 +530,29 @@ findSource nixosChannels channel source =
...
+
+        asFlakeSourceLink flakeUrl_ value =
+            let
+                looksLikePath v =
+                    not (String.isEmpty v)
+                        && not (String.contains " " v)
+                        && not (String.contains "," v)
+
+                ref =
+                    Maybe.withDefault "HEAD" source.flakeRevision
```
raboof, line 542: does `HEAD` typically point to the main branch? neat, wasn't aware of that

frontend/src/Page/Options.elm
```diff
@@ -529,6 +530,29 @@ findSource nixosChannels channel source =
...
+
+                positionWithLine =
+                    cleanPosition value |> String.replace ":" "#L"
+            in
+            if looksLikePath value && String.startsWith "https://github.com/" flakeUrl_ then
```
raboof, line 547: how does this deal with non-GitHub forges? I don't have a particular example handy though

raboof (commented):
Still seems somewhat GitHub-centric, but an improvement in any case!

raboof (approved):

frontend/src/Page/Options.elm
aciceri: On the production index, every option doc with a non-null `option_source` has `flake_resolved.type = github` unfortunately, haven't investigated why/how it's built.

frontend/src/Page/Options.elm
aciceri: Exactly

PR #1227, opened 2026-04-25
feat(flake-info): handle compound licenses from nixpkgs

KiaraGrouwstra:
This extends the `License` enum to deserialize the new `licenseType`-tagged structures, recursively flattens them into leaf licenses for the search index, and exposes a `package_license_expression` field carrying the SPDX-style [expression string](https://spdx.github.io/spdx-spec/v3.0.1/annexes/spdx-license-expressions/) (AND, OR, WITH, PLUS) when compound licenses are present (see NixOS/nixpkgs#468378).

Bump import index version to 47 for the new field.

Note that a potential follow-up would be to compound-proof our license filter, which under this PR just matches on any package containing a license in its expression.

### DIFF
--- a/flake-info/src/data/export.rs
+++ b/flake-info/src/data/export.rs
@@ -48,11 +48,111 @@ impl From<import::License> for License {
                 fullName,
                 shortName,
                 url,
+                ..
             } => License {
                 url,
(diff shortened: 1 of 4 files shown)


raboof (commented):
This should update the `import` version, right? (I suppose that was lost in a rebase)

raboof (approved):

PR #1230, opened 2026-04-25
feat(frontend): link Functions to noogle.dev

### LINKED ISSUE #1229: document functions
nix functions are currently documented at @hsjobeki's [noogle.dev](https://noogle.dev/).
given how useful such info is to nix users, it would be desirable to formally offer documentation on these.

KiaraGrouwstra:
As a stop-gap for documenting Nix functions, add a `Functions` entry to the navigation pointing at <https://noogle.dev>, alongside the existing `NixOS Wiki` external link.

Closes #1229.

### DIFF
--- a/frontend/src/Main.elm
+++ b/frontend/src/Main.elm
@@ -484,7 +484,9 @@ viewNavigation route =
             , ( Route.Options searchArgs, text "Options" )
             , ( Route.Flakes searchArgs, text "3rd-party Flakes" )
             ]
-        ++ [ li [] [ a [ href "https://wiki.nixos.org" ] [ text "NixOS Wiki" ] ] ]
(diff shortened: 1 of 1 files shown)


hsjobeki (approved):
Fyi: Currently improving the toolchain for retrieving those documentation. Would be happy to offer those as a library, but it needs some more dedicated time. I think the stop-gap is a nice ux improvement for now. We should just be clear that this is only a temporary solution.

raboof (approved):
seems reasonable to me.

thinking of the target audience of the search, beginning users might not need it yet, but part of the nice thing of Nix is that you can gradually start doing more complicated things, so having the function reference at hand seems helpful.

I understand there's people planning some kind of unified search spanning packages, modules, 3rd party flakes, nix.dev documentation etc - indeed this'd be a temporary thing until that happens.

PR #1231, opened 2026-04-25
feat(frontend): add `j`/`k` keybinds to move between search results

### LINKED ISSUE #1086: Feature request: J/K keybinds to move focus between results
### Feature Request

It would be nice and ergonomic for more keyboard-centric folks out there (which I can imagine a tech community like our have a lot of them!), to have a better way of moving around search results. 

I'd like to see `J` and `K` keybinds to move up or down the focus of *just* the search results (not all the links like `Tab` does).

We already have `Ctrl+K` and `/` keybinds for jumping to search box, so hopefully it won't be too hard to add new ones.

### Prior Art

- Github has `J`/`K` keybinds that work both on issue/PR views, and on Code view as well.
- DuckDuckGo has `J`/`K` keybinds on the search results view.

KiaraGrouwstra:
Adds vim-style navigation (matching the keybinds used by GitHub and DuckDuckGo) for moving between result items on the package, options and flake search pages. Pressing `j` selects the next result, `k` the previous, and `Enter` activates the selected result the same way a mouse click on its title would (toggling its expanded view).

Selection is tracked via a `result-selected` class rather than browser focus, since the result `<li>`s are not focusable by default and the global `*:focus { outline-width: 0 }` rule would hide any focus ring. The class reuses the existing `outline-visible` mixin so the active result gets the same blue outline used for `:focus-visible`.

The handler ignores key events from inputs and any modifier combinations, so existing shortcuts (e.g. `Ctrl+K`, `/`) and normal typing in the search box are unaffected.

Closes #1086.

### DIFF
--- a/frontend/src/index.js
+++ b/frontend/src/index.js
@@ -40,3 +40,102 @@ document.addEventListener("DOMContentLoaded", function () {
         });
     }
 });
+
+// `j` / `k` move the selection between search result items, `Enter`
+// activates the selected one. Selection is tracked with a class instead
+// of browser focus, since the result `<li>`s are not focusable by default
+// and the global `*:focus { outline-width: 0 }` rule would hide any focus
+// ring anyway.
+const SELECTED_CLASS = "result-selected";
+let selectedResultId = null;
+
+function isTypingTarget(target) {
+    return (
+        target &&
+        target.matches &&
+        target.matches(
+            "input, textarea, select, [contenteditable=true], [contenteditable='']",
+        )
+    );
+}
+
+function applySelection() {
+    document.querySelectorAll("." + SELECTED_CLASS).forEach((el) => {
+        if (el.id !== selectedResultId) {
+            el.classList.remove(SELECTED_CLASS);
+        }
+    });
+    if (!selectedResultId) {
+        return;
+    }
+    const el = document.getElementById(selectedResultId);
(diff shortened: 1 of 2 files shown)


frontend/src/index.js
```diff
@@ -33,3 +33,102 @@ document.addEventListener("DOMContentLoaded", function () {
...
+    const el = document.getElementById(selectedResultId);
+    if (el && !el.classList.contains(SELECTED_CLASS)) {
+        el.classList.add(SELECTED_CLASS);
+    }
+}
+
+// Elm's VDOM diff replaces the `class` attribute on re-render (e.g. when
+// toggling a result's expanded view), which would otherwise wipe our
+// `result-selected` class. Re-apply it whenever the DOM changes.
+new MutationObserver(applySelection).observe(document.body, {
```
raboof, line 80: Does this suggest it make more sense to implement this functionality in Elm instead of in js?

KiaraGrouwstra: in this case that [seems worse](https://github.com/NixOS/nixos-search/compare/feat-jk-result-navigation...feat-jk-result-navigation-elm) tbh

raboof (approved):

KiaraGrouwstra:
i'm still wondering how to perhaps better communicate the addition of this feature to users

PR #1235, opened 2026-04-26
fix(flake-info): tighten home-manager flake detection

### LINKED ISSUE #1234: github:fort-nix/nix-bitcoin can't be imported
seen earlier in #1175 but now I can actually reproduce locally as well

KiaraGrouwstra:
Detecting home-manager flakes by `modules/modules.nix` alone produced false positives for unrelated flakes that happen to ship a similarly named file (e.g. `nix-bitcoin`). Once a non-HM flake matched, the subsequent `import` of `modules/lib/stdlib-extended.nix` aborted the whole evaluation, so packages and options for those flakes also failed to index.

Also require `modules/lib/stdlib-extended.nix` to be present, which is specific to home-manager and is the helper `readHomeManagerOptions` already imports.

Closes #1234.

### DIFF
--- a/flake-info/assets/commands/flake_info.nix
+++ b/flake-info/assets/commands/flake_info.nix
@@ -518,9 +518,17 @@ rec {
   options = readFlakeOptions;
   home-manager-options =
     let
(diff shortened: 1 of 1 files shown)


raboof (approved):
thanks!

PR #1237, opened 2026-04-27
feat(frontend): add debounced typeahead suggestions to search input

KiaraGrouwstra:
## Summary
- New `Search.Typeahead` module: 150 ms debounced, stale-token drop, hits the existing `*.edge` ngram subfields with `size: 8`. Suggestions render as anchor links that navigate straight to the package/option detail.
- Honors `navigator.connection.saveData` (and `effectiveType in {slow-2g, 2g}`) - the dropdown stays disabled on metered/slow connections and the page falls back to the original submit-on-enter UX.
- Closes the dropdown on submit, Escape, and blur (200 ms grace so a click on a suggestion still navigates). Re-focusing the input re-shows cached suggestions.
- The Flakes page keeps it disabled (its aggregate index is not channel-shaped).

No backend or mapping changes - reuses existing edge subfields. If ranking quality turns out to be poor we can follow up with a `search_as_you_type` field; that would require an index version bump.

## Test plan
- `npm run prod` builds clean.
- Type into the Packages search bar - dropdown appears after pause, click navigates.
- Same on the Options page - suggestions are options, not packages.
- Press Enter / Escape / click outside - dropdown closes.
- Refocus the input - cached dropdown re-appears.
- Toggle Chrome DevTools "Slow 3G"/"Save-Data" - dropdown stays hidden.
- Flakes page - no dropdown, behaves as before.

### DIFF
--- a/frontend/src/Main.elm
+++ b/frontend/src/Main.elm
@@ -55,6 +55,7 @@ type alias Flags =
     , elasticsearchUsername : String
     , elasticsearchPassword : String
     , nixosChannels : Json.Decode.Value
+    , saveData : Bool
     }
 
 
@@ -65,6 +66,7 @@ type alias Model =
     , defaultNixOSChannel : String
     , nixosChannels : List NixOSChannel
     , page : Page
+    , typeaheadEnabled : Bool
     }
 
 
@@ -107,6 +109,7 @@ init flags url navKey =
             , nixosChannels = nixosChannels.channels
             , page = NotFound
             , route = Route.Home
+            , typeaheadEnabled = not flags.saveData
             }
     in
     changeRouteTo model url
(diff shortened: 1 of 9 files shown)


raboof:
there seems to be an extra bulleted list under the search box that pops in and out of view?

raboof:
While I think this is neat, I'm a bit concerned that this will seriously increate the load on our ES instance, which already seems to be struggling sometimes. What could we do to monitor/compensate for that?

KiaraGrouwstra:
maybe before this change to lighten ES load i can look into statically pre-loading light result sets on connections that are not metered (or marked data saver for that matter)

raboof (approved):
let's see!

PR #1244, opened 2026-04-30
for home-manager, filter displayed options by branches deemed to match the selected nixpkgs one

### LINKED ISSUE #1242: Hardcoded HM branch causes wrong results for stable channel
i just tripped over the options search showing wrong ( nonexisting)  options for the stable branch for hm options specifically. 

i think the problem stems from hardcoding the branch in  ‎frontend/src/Page/Options.elm .

my naive understanding is that changing the branch from
```
        homeManagerUrlPrefix =
            "https://github.com/nix-community/home-manager/blob/master/"
```
to
```
        homeManagerUrlPrefix =
            "https://github.com/nix-community/home-manager/blob/" ++ branch ++ "/"
```

should probably fix this.

KiaraGrouwstra:
to translate the spirit of the UI without further complicating it for the time being, implicitly maps nixpkgs branch `nixos-unstable` to HM branch `master,` and stable branches to corresponding stable branches.

Closes #1242.

### DIFF
--- a/flake-info/src/commands/nixpkgs_info.rs
+++ b/flake-info/src/commands/nixpkgs_info.rs
@@ -181,15 +181,25 @@ pub fn get_nixpkgs_services(nixpkgs: &Source) -> Result<Vec<NixpkgsEntry>> {
     Ok(attr_set.into_iter().map(NixpkgsEntry::Service).collect())
 }
 
-/// Home-manager flake reference used to evaluate HM options alongside
-/// each nixpkgs channel import.
-const HOME_MANAGER_FLAKE_REF: &str = "github:nix-community/home-manager";
+/// Home-manager flake reference matching a given nixpkgs channel. Stable
+/// nixpkgs channels (`nixos-XX.YY`) get the corresponding `release-XX.YY`
+/// branch in `nix-community/home-manager`; `nixos-unstable` gets `master`.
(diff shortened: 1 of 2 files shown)


version.nix
```diff
@@ -2,7 +2,7 @@
   /**
     Backend index version used by import jobs when writing data to Elasticsearch
   */
-  import = "47";
+  import = "48";
```
raboof: this doesn't change the index schema, does it?

KiaraGrouwstra: thanks, now split out the front-end bits to #1250

raboof: Are you sure a index version update is even needed? is there any regression when we apply the frontend changes without applying the backend changes?

KiaraGrouwstra: whoops, collapsed them back

KiaraGrouwstra:
test case: `programs.codex.rules` should be present under HM options only for unstable (as of writing)

raboof (approved):
gotcha - so https://69f4b00a2d903e258b33401a--nixos-search.netlify.app/options?channel=25.11&query=codex&show=home-manager-option:programs.codex.rules&source=home_manager now 'correctly' links to a page that doesn't have it, and we expect this to be fixed after the next import, right? (which might be a while since the current import will stay cached until the channel is bumped)

KiaraGrouwstra:
yeah - see the now-dynamic home-manager version in the import script.

Issue #1210, opened 2026-04-16
Feedback on recent UI changes for the options search


SuperSandro2000:
Recently the Options Search page has seen some changes and my monk immediately found this unordered list of findings. Due to me not having any idea about Elm I am unable to contribute any fixes for those things, unfortunately.

- The checkboxes are unaligned 

<img width="351" height="186" alt="Image" src="https://github.com/user-attachments/assets/823c781e-c6e7-4eb4-9466-c98615bdf921" />

- When changing the checkboxes the border becomes wider and immediately jumping back

<img width="524" height="193" alt="Image" src="https://github.com/user-attachments/assets/83990556-61c0-474c-8ef1-7a7656c3dad7" />

- The NixOS labels exist on all options except 18. I would just remove them and only add them for modular services. This would also reduce visual noise on the page.

<img width="460" height="342" alt="Image" src="https://github.com/user-attachments/assets/954ba221-994f-406c-9703-e009afffc706" />

- It is also a bit inconsistent that when unchecking Modular services on the left, the NixOS labels just vanish.

- The hint to search more precise is not very noticeable. In fact I only noticed it because the option listed is being shifted down and I wondered why that was the case

<img width="1320" height="418" alt="Image" src="https://github.com/user-attachments/assets/d6da84e4-9a4e-40e0-b99f-ba3f7af1ab3c" />

- Clicking on unstable again and again causes a refresh of the search result and also causes the border getting bigger and jumping back I described above

<img width="692" height="396" alt="Image" src="https://github.com/user-attachments/assets/f78f998b-4655-461d-b49f-c1eaa1bbe42d" />

- Linkifying the options is a good idea in itself but now copying them is really hard

<img width="493" height="181" alt="Image" src="https://github.com/user-attachments/assets/c1c17ded-cbfd-4752-ba2e-ef331b97e999" />

- Clicking on the highlighted preStart causes the filter on the left to be reset (everything is checked again) 

<img width="747" height="593" alt="Image" src="https://github.com/user-attachments/assets/c9ea30e3-e4cb-419c-b243-f2b84b533a2d" />

KiaraGrouwstra:
planning to look into this after #1202 so as to reduce rebasing

KiaraGrouwstra:
> It is also a bit inconsistent that when unchecking Modular services on the left, the NixOS labels just vanish.

[...]

PR #1246, opened 2026-04-30
feat(frontend): tab Options page sources, opt into ES request_cache

KiaraGrouwstra:
switch option categories from checkboxes to tabs to improve cacheability and decrease visual clutter.

image

Replaces the multi-checkbox `includedOptionSources : Set String` with a single `activeOptionSource : OptionSource`. The Options page now shows one source at a time per tab.
Each tab issues a single ES query with a fixed `term: { type: <one> }` filter, so the request body is byte-stable across users on identical input. With the new `?request_cache=true` opt-in (added on the new `Search.makeRequestTask` Task-based path used by Options), repeated queries become near-free shard cache hits rather than per-user noise.

Compared to the previous "single query with `terms` array" shape:

- Up to 7 distinct cache keys per popular query (one per checkbox combo) collapses to 1 per tab.
- Per-result source badges become redundant (one tab = one docType) and are dropped.

Each tab carries a hit-count badge. The active tab's count comes from the in-flight query result; inactive tabs are populated by a `size: 0` count query per tab fired alongside the main request. `size: 0` is the workload ES's `request_cache` was designed for, so once a popular query has been seen on any source the count queries cost almost nothing. Counts render as `1234`, `1.2k`, or `10k+` (ES caps `total.value` at 10000 unless we ask otherwise, so don't pretend the larger numbers are exact). Counts clear on every fresh load so stale badges don't linger; inactive errors are swallowed silently since the badge is a hint, not a result.

The new URL shape is `?source=<id>` (omitted for the default `nixos` tab).

Tackles https://github.com/NixOS/nixos-search/pull/1237#issuecomment-4351203986's concern regarding load on Elastic Search, and #1210's suggestion to return to a pre-badge uncluttered UI.

### DIFF
--- a/frontend/src/Main.elm
+++ b/frontend/src/Main.elm
@@ -3,6 +3,7 @@ module Main exposing (Flags, Model, Msg, Page, main)
 import Browser
 import Browser.Dom
 import Browser.Navigation
+import Dict
 import Html
     exposing
         ( Html
@@ -39,7 +40,6 @@ import Search
         , decodeNixOSChannels
         , defaultFlakeId
         )
-import Set
 import Shortcut
 import Task
 import Url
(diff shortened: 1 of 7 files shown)


raboof (approved):

raboof:
looks good on netlify preview

Issue #736, opened 2024-03-08
[Regression] Text exceeds border on mobile devices


Aleksanaa:
![Screenshot_20240308-134014_Firefox](https://github.com/NixOS/nixos-search/assets/42209822/47bd7141-4575-4e75-824f-d39ed3788c19)

As the screenshot shows.

PR #1255, opened 2026-05-03
fix(layout): better handling for narrow screens

cyrneko:
Previously, on narrow screens of e.g phones, the site did not adjust beyond a `max-width` of 760px, causing overflows like in #736 and generally making it hard to navigate.

This simple media query fixes that, and also expands the sidebar to full-width in such a case.

The only downside is that the user now needs to scroll past the filters and such.

### before

image

### after
image


raboof (approved):
nice!

PR #1262, opened 2026-05-04
feat(frontend): static-asset-backed type-ahead for options (modular services, home-manager)

KiaraGrouwstra:
This change implements type-ahead backed by static assets for the categories small enough to make that viable: options (for now) for modular services and home-manager.

This works using a one-time JSON fetch from `/autocomplete/<category>-<channel>.json`, after which the browser does matching client-side with multi-word ranking.

Metered connections (`cellular`) and data saver mode (`saveData`) are checked so as to opt into pre-loading such static assets outside of those scenarios.

The aforementioned type-ahead based on Elastic Search (rather than static assets), as proposed in #1237 earlier, may be added [as a follow-up](https://github.com/NixOS/nixos-search/compare/typeahead-ES).
For clients not opting into the static assets those may then function as a fallback.

Generation of the static assets containing the option names is done in a new step in CI job `frontend.yml`, falling back to `[]` on failure.
This way, we can grab the option names needed to generate the assets upon builds of the front-end, so we don't need to evaluate nixpkgs for this in e.g. our ES import job that runs every two hours.

Supersedes #1237.

### DIFF
--- a/frontend/scripts/serve-local.mjs
+++ b/frontend/scripts/serve-local.mjs
@@ -1,0 +1,63 @@
+#!/usr/bin/env node
+// Local static server that mirrors the two `netlify.toml` redirect rules:
+//   /backend/* → proxy to bonsai (Elasticsearch)
+//   /*         → dist/index.html (SPA fallback)
+import http from 'node:http';
+import https from 'node:https';
+import fs from 'node:fs';
+import path from 'node:path';
+
+const DIST = path.resolve(process.argv[2] ?? 'dist');
+const BONSAI = 'nixos-search-7-1733963800.us-east-1.bonsaisearch.net';
+const PORT = process.env.PORT ?? 8080;
+
+const MIME = new Map([
+  ['.html', 'text/html; charset=utf-8'],
+  ['.js', 'application/javascript'],
+  ['.css', 'text/css'],
+  ['.json', 'application/json'],
+  ['.svg', 'image/svg+xml'],
+  ['.png', 'image/png'],
+  ['.ico', 'image/x-icon'],
+  ['.woff2', 'font/woff2'],
+  ['.woff', 'font/woff'],
+  ['.txt', 'text/plain'],
+]);
+
+http.createServer((req, res) => {
+  const reqUrl = new URL(req.url, `http://localhost:${PORT}`);
+
+  if (reqUrl.pathname.startsWith('/backend/')) {
+    const targetPath = reqUrl.pathname.slice('/backend'.length) + reqUrl.search;
+    const opts = {
+      hostname: BONSAI,
+      path: targetPath,
+      method: req.method,
+      headers: { ...req.headers, host: BONSAI },
+    };
+    // Avoid compressed responses we'd need to decompress before forwarding.
+    delete opts.headers['accept-encoding'];
+    const proxy = https.request(opts, (proxyRes) => {
+      res.writeHead(proxyRes.statusCode, proxyRes.headers);
+      proxyRes.pipe(res, { end: true });
+    });
+    proxy.on('error', (e) => { res.writeHead(502); res.end(String(e)); });
+    req.pipe(proxy, { end: true });
+    return;
+  }
+
+  let filePath = path.join(DIST, reqUrl.pathname);
+  try {
+    if (fs.statSync(filePath).isDirectory()) filePath = path.join(filePath, 'index.html');
+  } catch {
+    filePath = path.join(DIST, 'index.html');
+  }
+  if (!fs.existsSync(filePath)) filePath = path.join(DIST, 'index.html');
+
+  const ext = path.extname(filePath);
+  res.writeHead(200, { 'Content-Type': MIME.get(ext) ?? 'application/octet-stream' });
+  fs.createReadStream(filePath).pipe(res, { end: true });
+}).listen(PORT, () => {
+  console.log(`Serving ${DIST} at http://localhost:${PORT}`);
+  console.log(`Proxying /backend/* → https://${BONSAI}`);
+});
--- a/frontend/scripts/serve-local.mjs
+++ b/frontend/scripts/serve-local.mjs
@@ -55,7 +55,17 @@ http.createServer((req, res) => {
   if (!fs.existsSync(filePath)) filePath = path.join(DIST, 'index.html');
 
   const ext = path.extname(filePath);
-  res.writeHead(200, { 'Content-Type': MIME.get(ext) ?? 'application/octet-stream' });
+  const headers = { 'Content-Type': MIME.get(ext) ?? 'application/octet-stream' };
+  if (ext === '.html' && req.headers['save-data'] === 'on') {
+    fs.readFile(filePath, 'utf8', (err, data) => {
+      if (err) { res.writeHead(500); res.end(String(err)); return; }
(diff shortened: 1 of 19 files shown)


Eveeifyeve (approved):
LTGM! Tested on some autocompletes in home-manager, nixos and modular services in the deployment and Diff LTGM.

frontend/scripts/serve-local.mjs
```diff
@@ -0,0 +1,63 @@
...
+// Local static server that mirrors the two `netlify.toml` redirect rules:
+//   /backend/* → proxy to bonsai (Elasticsearch)
+//   /*         → dist/index.html (SPA fallback)
+import http from 'node:http';
+import https from 'node:https';
+import fs from 'node:fs';
+import path from 'node:path';
+
+const DIST = path.resolve(process.argv[2] ?? 'dist');
+const BONSAI = 'nixos-search-7-1733963800.us-east-1.bonsaisearch.net';
```
raboof, line 11: could this load /frontend/config/webpack.dev.js or does that make no sense?

frontend/src/index.js
```diff
@@ -5,6 +5,14 @@ require("elm-keyboard-shortcut");
...
 const { Elm } = require("./Main");
 
+const connection =
+    navigator.connection ||
+    navigator.mozConnection ||
+    navigator.webkitConnection;
+const saveData =
+    (connection && connection.saveData) ||
+    (connection && connection.type === "cellular");
+
```
raboof: 'connection' is 'undefined' for me on Firefox - would there be an equivalent there?

raboof (commented):
Very cool!

I'm afraid other merges have caused some conflicts.

I don't actually see the feature on https://6a003a6882572ceed42cf583--nixos-search.netlify.app/options , neither on Firefox nor on chromium - is that expected? I don't see it request the index.

Eveeifyeve (changes requested):
Actually I agree with @raboof, There is still an issue with nixos type ahead options. Home-manager and modular services is fine, just nixos. I would expect support for type-ahead for nixos as well.

frontend/src/Search/Typeahead.elm
```diff
@@ -0,0 +1,405 @@
...
+        docType =
+            case category of
+                "home_manager" ->
+                    "home-manager-option"
+
+                "modular_service" ->
+                    "service"
+
+                _ ->
+                    "option"
```
Eveeifyeve, line 310: Does this actually work for nixos?

frontend/src/index.js
KiaraGrouwstra: thanks, good catch, turns out they removed it over privacy concerns. i added a commit to communicate info on data saver mode from the server now.

KiaraGrouwstra:
@Eveeifyeve: NixOS options are currently omitted as they'd make for way bigger static assets:

> static assets for the categories small enough to make that viable: options (for now) for modular services and home-manager.

this is particularly relevant given firefox appears to have dropped info on metered connections (`navigator.connection`) over fingerprinting concerns, meaning that for mobile devices all we'd have to go by would be explicit data saver mode.
that kind of sucks, because it means we should definitely keep eagerly loaded assets somewhat modest. alternatives would be opt-ins and type-ahead powered by requests to the ES server.

the latter i want to add in a follow-up PR.

frontend/src/Search/Typeahead.elm
KiaraGrouwstra: good catch - not with the static assets we're limited to right now.
we could leave it for the follow-up maybe - this could be a more sane default than the others.

raboof:
> for mobile devices all we'd have to go by would be explicit data saver mode. that kind of sucks, because it means we should definitely keep eagerly loaded assets somewhat modest.

Maybe we could look at the screen resolution and make assumptions based on that?

> alternatives would be opt-ins and type-ahead powered by requests to the ES server.
> 
> the latter i want to add in a follow-up PR.

I'm still concerned about the added load that'd put on the ES server.

frontend/scripts/serve-local.mjs
KiaraGrouwstra: the intent with this one is to use the corpus in `dist/` generated thru github actions for prod (deployed to netlify), while webpack would just uses the dev bundle.
duplication seems not really bad enough to factor stuff out here either.

KiaraGrouwstra:
> I'm still concerned about the added load that'd put on the ES server.

load has seemed at 30% ish so far - i think given a reasonable debounce time (only try when the user actually finished typing their word) this may well work out ok. moreover, given the type-ahead means less info to send over the wire than a proper search, while the info may suffice to help users narrow hits down to relevant sub-modules sooner, i think this may not turn out all negative - but perhaps that's a discussion more relevant for such a follow-up.

> Maybe we could look at the screen resolution and make assumptions based on that?

phones can also be on wifi, while resizing a window on desktop wouldn't make a connection metered. not a great proxy measure unfortunately...

KiaraGrouwstra:
what's the status on this one? should i rebase yet?

KiaraGrouwstra:
first gonna handle #1312

Mic92:
@KiaraGrouwstra I spent a little bit time on improving the search scoring, it's still not perfect because we don't have any popularity index but it improves direct hits quit a bit i.e. you can now look for `gcc` and it actually returns gcc first: https://gist.github.com/Mic92/ff0407e1f090d12ef3abd5b726f34d01

Mic92:
What I think worth trying is to use nix-eval-jobs with `--no-instantiate` and `--show-input-drvs` to compute a dependency graph and than compute a popularity score based on the page rank algorithm. (so the more a dependency is used, the more popular it gets.

eveeifyeve:
> than compute a popularity score based on the page rank algorithm. (so the more a dependency is used, the more popular it gets.

Isn't this a form of telemetry? I am just worried about the foundation being unsure about GDPR compliance if this is the case.

KiaraGrouwstra:
isn't GDPR just about personally identifiable (be it in isolation or once combined) info? cuz would 'package x got downloaded' actually qualify there?

KiaraGrouwstra:
> first gonna handle #1312

with that down, going by [cpu load](https://github.com/NixOS/infra/pull/1074#issuecomment-4914340962) i think we're good still - i'd argue we could try the type-ahead and revert if it turns out to overload the instance. @raboof?

eveeifyeve:
> isn't GDPR just about personally identifiable (be it in isolation or once combined) info? cuz would 'package x got downloaded' actually qualify there?

I am not sure, as I haven't gotten around the to doing GDPR stuff before.

KiaraGrouwstra:
on load stats, i added a [grafana dashboard](https://grafana.nixos.org/d/cimnqkt/elasticsearch-load-nixos-search)

eveeifyeve:
It would be good to support something like if someone searches "docker options" to have some good search suggestions.

Mic92:
> > than compute a popularity score based on the page rank algorithm. (so the more a dependency is used, the more popular it gets.
> 
> Isn't this a form of telemetry? I am just worried about the foundation being unsure about GDPR compliance if this is the case.

No. This is not even using downloads, which would require some more machinery, I wouldn't want to build. This is how nixpkgs internally is structured. But even download statistics without IPs are not related to GDPR.

raboof (approved):

PR #1269, opened 2026-05-08
update lock file to account for switch to explicit flake pins

KiaraGrouwstra:
follow-up to #1265, whoops

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -44,9 +44,10 @@
         "type": "github"
       },
       "original": {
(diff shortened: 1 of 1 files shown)


raboof (approved):
no objection, but isn't this automatically picked up with the generated 'bump' PR's?

KiaraGrouwstra:
i imagine it should be - yes. until then tho, this prevents irrelevant diffs when changing other stuff.

PR #1276, opened 2026-05-10
don't add `channel` query param in flakes section that didn't exist already

### LINKED ISSUE #1263: Flakes - Remove Channel URL Parameter
On https://search.nixos.org/flakes, clicking on buttons will append a `channel` url parameter (i.e. `channel=25.11`)

It is irrelevant in flakes context, and should not be **appended**

However, switching from https://search.nixos.org/packages?channel=unstable&query=hello / https://search.nixos.org/options?channel=unstable&query=hello to https://search.nixos.org/flakes should not remove the `channel` url parameter (in case the user switch back to https://search.nixos.org/packages / https://search.nixos.org/options afterwards)

KiaraGrouwstra:
Prevents having searches in the flakes section for no reason add `channel` query params.

In other words, the URL when issuing a search from the flakes section, will be like:

- before: https://search.nixos.org/flakes?channel=25.11&query=nix
- after: http://localhost:3000/flakes?query=nix

Closes #1263.

### DIFF
--- a/frontend/src/Main.elm
+++ b/frontend/src/Main.elm
@@ -283,7 +283,7 @@ changeRouteTo currentModel url =
                         _ ->
                             Nothing
             in
-            Page.Packages.init searchArgs currentModel.defaultNixOSChannel currentModel.nixosChannels modelPage
+            Page.Packages.init searchArgs currentModel.defaultNixOSChannel currentModel.nixosChannels True modelPage
                 |> updateWith Packages PackagesMsg model
                 |> avoidReinit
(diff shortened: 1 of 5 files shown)


raboof (approved):
Hmm, so as long as there's no channel the flakes tab doesn't add it anymore (good), but once a channel has been selected the flakes tab does 'keep' it. I guess that makes sense (otherwise you'd lose the channel when switching tabs), though it's getting a bit hard to follow. I don't have a strong other proposal, though, so I guess this seems OK

PR #1277, opened 2026-05-12
strip out backward compatibility with `show` as query parameter, fixing a bug when switching option category

KiaraGrouwstra:
drops backward compatibility with the `show` query param.

this fixes an issue reported by @kranzes that makes the `show` fragment param into a query when switching option categories when using the ClearURLs extension on firefox v150:

image

not having managed to reproduce this without extension tho, that makes me feel somewhat hesitant on a path forward:

- reverting to the query param would make it harder to get cache hits
- breaking compatibility with the query param could break deep-linking predating the switch
- the issue reported here so far we've managed to manually reproduce only with ClearURLs so far

like all options there seem kinda imperfect, out of what i've come up with so far.


KiaraGrouwstra:
i'll maybe close until we find a path that seems not bad.

PR #1278, opened 2026-05-12
enhance(flakes): convert to standardized flake registry

### LINKED ISSUE #1183: enhance(flakes): convert to standardized flake registry
- **File** : `flake-registry.json` (like https://github.com/NixOS/flake-registry)
- **Format** : https://nix.dev/manual/nix/latest/command-ref/new-cli/nix3-registry.html#registry-format
- **Interface** : https://github.com/NixOS/nix/blob/master/src/libfetchers/include/nix/fetchers/registry.hh

We then would be able to use [`nix registry`](https://nix.dev/manual/nix/latest/command-ref/new-cli/nix3-registry.html), including [`nix registry add`](https://nix.dev/manual/nix/latest/command-ref/new-cli/nix3-registry-add) (which would make adding flakes easier, doable in CI using [`on.workflow_dispatch.inputs`](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onworkflow_dispatchinputs), and #1032 a non-starter), [`nix registry list`](https://nix.dev/manual/nix/latest/command-ref/new-cli/nix3-registry-list.html) (which would greatly help with #1181), but most importantly [`nix registry resolve`](https://nix.dev/manual/nix/latest/command-ref/new-cli/nix3-registry-resolve.html) (which would enable a great reduction of [flake-info](https://github.com/NixOS/nixos-search/tree/main/flake-info)'s LOC)

Malix-Labs:
CLOSES: #1183


README.md
```diff
@@ -91,23 +91,23 @@ You may need to manually edit `frontend/src/Search.elm` to use the right index.
 
 ## Adding flakes
 
-To add your own flakes to the search index edit [./flakes/manual.toml](./flakes/manual.toml), keeping the alphabetical ordering.
+To add your own flakes to the search index edit [./flakes/flake-registry.json](./flakes/flake-registry.json), keeping the alphabetical ordering.
 
-Possible types are `github`, `gitlab`, `sourcehut`, and `git` (which is the fallback for any kind of git repository but requires to set a revision key manually as of now).
+This registry currently contains `github`, `gitlab`, and `sourcehut` entries.
```
Eveeifyeve, line 96: Would it be good to reference https://nix.dev/manual/nix/2.25/command-ref/new-cli/nix3-flake.html?highlight=flake%20refere#types instead here?

Malix-Labs: The intended workflow would be to abstract that entirely through `nix registry add`

The readme would get simplified as a result

Please also see https://github.com/NixOS/nixos-search/issues/1183#issuecomment-4464715712 (and you can subscribe to this issue too, since there will be long running developments)

Eveeifyeve: Is this regarding to https://github.com/NixOS/nix/pull/15660? Because if it is, I might look at it and review.

Malix-Labs: Yep! and not only that, there is also NixOS/nix#4780, and perhaps also other but related elements I've not reported

Do you have a synchronous chat (discord message or call, matrix messages or call, SimpleX, Zulip, Signal, anything) where we could review / talk about it?

My discord is `malix_labs` and matrix is `@malix_off:matrix.org`

Eveeifyeve: > Do you have a synchronous chat (discord message or call, matrix messages or call, SimpleX, Zulip, Signal, anything) where we could review / talk about it?

Yes I do, I have both my discord which is `eveeifyeve` or my matrix `@eveeifyeve:matrix.org`. I don't mind both. UPDATE: I added you on both.

PR #1283, opened 2026-05-21
fix: devshell for darwin machines

eveeifyeve:
Fixes the issue with running devshell with apple-sdk on darwin machines.


raboof (approved):
OK if CI agrees I suppose ;)

/cc @sakompella

Issue #1281, opened 2026-05-15
Remove support for home-manager


eclairevoyant:
h-m is a separate project completely unaffiliated with nixpkgs/nixos and doesn't follow its basic coding standards. In fact in many cases it goes against the design of nixos.

Having it in the search is more noise and unnecessarily privileges one random project over any others, making it look official. (Even modular services should be marked as "experimental" to make it clear that it's not officially in use as of now.)

KiaraGrouwstra:
i'm seeing some mixed reactions here, which maybe raises the question what considerations those here saw.

mine i guess were maybe somewhat simple: i'd like to see this project grow in functionality, while i see in practice people seem to depend on more than just NixOS proper.
this wasn't so much meant as a knock on other projects like `nix-darwin` or `hjem-rum`, just gradual steps in trying to see how we may better serve nix users' needs.

i'll admit as a nixos user i've personally found it hard to manage a usable workstation without resorting to use of home-manager so far, so that might bias me here - tho perhaps some of the others who reacted here felt similar.

> (Even modular services should be marked as "experimental" to make it clear that it's not officially in use as of now.)

fair!

> h-m is a separate project completely unaffiliated with nixpkgs/nixos

to be fair, i've felt under the impression that nixos.org has been less nixos-only, but rather about nix as well, which in case one would extend such reasoning to search.nixos.org definitely raises some valid questions about proper scope here.

> Having it in the search is more noise [...]

this sounds like a UX thing - which is good feedback, tho then there might be multiple ways to approach this.
i'm not sure what the solution here is, and i might be making things less consistent by blurring the boundary between the main options page vs the 'third-party flakes' one - and maybe we might be better off changing that boundary, if not finding further inspiration at <https://search.nüschtos.de/>.

> Having it in the search [...] unnecessarily privileges one random project over any others

if we'd consider adding further projects, i wonder if the concern might gradually alleviate - such an opt-in fashion would be not unlike what had been used here for the flakes page for longer already.

> making it look official

this sounds like a matter of setting expectations - i wonder if just like with modular services and marking as experimental, marking this particular repo as being in the `nix-community` org might help prevent at least some such misunderstandings?

now, to be fair, while i've had some activity at this repo lately, many the design decisions here also feel like subject to the documentation team (/cc @hsjobeki), whose [document](https://github.com/NixOS/nix.dev/blob/master/maintainers/vision.md) i might have read, tho they're probably a better arbiter of interpreting such overall strategy.

[...]

PR #1288, opened 2026-05-26
mark Home Manager as part of `nix-community`

KiaraGrouwstra:
we could revert this when the project is officially marked as part of the NixOS org (see #1281), but until then this may set more realistic expectations


raboof (approved):

raboof:
This is a nice improvement. I'd still like a oneliner saying what home-manager is and linking to https://github.com/nix-community/home-manager/, but that can be a separate improvement.

Malix-Labs:
Related : #1397

PR #1289, opened 2026-05-26
mark modular services as experimental

KiaraGrouwstra:
Mark modular services in the options tab as experimental, as suggested in #1281.

### DIFF
--- a/frontend/src/Page/Options.elm
+++ b/frontend/src/Page/Options.elm
@@ -229,6 +229,16 @@ viewSourceTab activeSource count source =
         id =
             Route.optionSourceId source
 
+        labelChildren =
+            text (Route.optionSourceLabel source)
+                :: (case source of
+                        Route.ModularServiceOptions ->
+                            [ span [ class "label label-info experimental-badge" ] [ text "Experimental" ] ]
+
+                        _ ->
+                            []
+                   )
+
         badge =
             case count of
                 Just n ->
@@ -247,7 +257,7 @@ viewSourceTab activeSource count source =
             , href "#"
             , Html.Events.onClick (SearchMsg (Search.SetActiveOptionSource source))
             ]
-            (span [] [ text (Route.optionSourceLabel source) ] :: badge)
+            (span [ class "source-label" ] labelChildren :: badge)
         ]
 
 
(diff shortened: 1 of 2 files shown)


raboof (approved):
Sounds good.

I could also see only showing this category when the 'unstable' channel is selected.

It would still be nice to link to https://nixos.org/manual/nixos/unstable/#modular-services somewhere. (also shouldn't that be in the nixpkgs manual rather than the nixos-manual? I thought kinda the point of modular services was that they're not NixOS-specific :)

KiaraGrouwstra:
> It would still be nice to link to https://nixos.org/manual/nixos/unstable/#modular-services somewhere.

We do this today! It has been linked at options details' entry:

> **About**: [What are modular services?](https://nixos.org/manual/nixos/stable/#modular-services)

> (also shouldn't that be in the nixpkgs manual rather than the nixos-manual? I thought kinda the point of modular services was that they're not NixOS-specific :)

Indeed! This point was raised earlier at https://github.com/NixOS/nixpkgs/issues/432550#issuecomment-4255111511, and I've worked on this a bit at NixOS/nixpkgs#519015.

> I could also see only showing this category when the 'unstable' channel is selected.

To be fair, the feature has been available on stable, tho it's maybe unfortunate that the name might give the impression it's already in the ballpark of usefulness as NixOS services, of which we have various orders of magnitude more today.

I feel like a way to communicate that would be to have its limited set of options eagerly loaded such that people would already see there's very little there yet today. (I'd see #1262 as the first step toward achieving that.)

raboof:
> > It would still be nice to link to https://nixos.org/manual/nixos/unstable/#modular-services somewhere.
> 
> We do this today! It has been linked at options details' entry:
> 
> > **About**: [What are modular services?](https://nixos.org/manual/nixos/stable/#modular-services)

Yeah that's already good. I think it might make sense to have it above rather than 'inside' the search results at some point.
 
> > (also shouldn't that be in the nixpkgs manual rather than the nixos-manual? I thought kinda the point of modular services was that they're not NixOS-specific :)
> 
> Indeed! This point was raised earlier at [NixOS/nixpkgs#432550 (comment)](https://github.com/NixOS/nixpkgs/issues/432550#issuecomment-4255111511), and I've worked on this a bit at [NixOS/nixpkgs#519015](https://github.com/NixOS/nixpkgs/pull/519015).

:+1:

PR #1290, opened 2026-05-26
chore: update nixos-infra flake input

eveeifyeve:
This flake update adds 26.05 as beta to packages and options.
Ref: https://github.com/NixOS/nixpkgs/issues/503391

A follow up will have to be done upon release, which I am happy to do.


yayayayaka (approved):
Literally one minute before pushing my own PR :laughing:

PR #1299, opened 2026-06-01
fix: switch logo workflow to create a pr

eveeifyeve:
Ref: https://github.com/NixOS/nixos-search/actions/runs/26699806552

---
> [!WARNING]
> Assisted by AI.

### DIFF
--- a/.github/workflows/switch-logo.yml
+++ b/.github/workflows/switch-logo.yml
@@ -1,6 +1,7 @@
 name: Switch Logo
 
 on:
+  workflow_dispatch:
   schedule:
     - cron: "0 0 31 5 *" # Change to pride logo. "At 00:00 on day-of-month 31 in May."
     - cron: "0 0 1 7 *"  # Switch back to normal logo. "At 00:00 on day-of-month 1 in July."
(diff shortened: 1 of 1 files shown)


.github/workflows/switch-logo.yml
```diff
@@ -1,36 +1,50 @@
 name: Switch Logo
 
 on:
+  workflow_dispatch:
```
raboof, line 4: this is not _super_ useful given the `if [ "$TODAY"` checks, but sure, might come in handy

.github/workflows/switch-logo.yml
```diff
@@ -1,36 +1,50 @@
...
+            pr_title="chore(logo): switch back to normal logo"
+            pr_body="Reverts the logo back to the standard variant after June."
           else
             echo "No action needed for logo today."
             exit 0
           fi
 
           git config --global user.name 'github-actions[bot]'
           git config --global user.email '41898282+github-actions[bot]@users.noreply.github.com'
+          git checkout -b "$branch"
```
raboof, line 42: (this will fail if we forget to remove last years' branch, OK I suppose)

raboof (approved):

.github/workflows/switch-logo.yml
Eveeifyeve: That's why I added the workflow_dispatch above.

PR #1311, opened 2026-06-05
for missing HM options link to the HM issue tracker

KiaraGrouwstra:
when there are no search results, we used to just refer people to the nixpkgs issue tracker.
this gates it to use the HM one in case of HM options.

/cc @Eveeifyeve


eveeifyeve:
Where is the web preview for this pr?

KiaraGrouwstra:
looks like https://github.com/NixOS/nixos-search/actions/runs/27031892636/job/79786054891 complains:

> Netlify credentials not provided, not deployable

/cc @raboof

this maybe raises the question when this regressed (?).
fwiw #1277 still got a preview.
more recent ones i haven't found any for just yet.

KiaraGrouwstra:
d'oh, looks like the auto-deploys need the secrets, which GHA only exposes for PRs made from the repo itself, not from forks.

raboof (approved):
LGTM but has a conflict

PR #1314, opened 2026-06-05
feat(frontend): copy confirmation toast and copy buttons for option values

ELHart05:
Two related improvements to the copy buttons.

1. A toast. Every copy button (package commands, option names) wrote to the clipboard with no feedback. They all go through the copyToClipboard port, so a "Copied to clipboard" toast is added to that one handler and shows only when the write succeeds. It uses the existing theme variables, respects prefers-reduced-motion, and has role="status" for screen readers.

2. More copy buttons. The options page only had a copy button on the option name. The Default, Example and Usage code blocks now have one too, Usage being the ready-to-paste configuration.nix snippet. They reuse the same port, so the toast covers them as well.

Tested against the production index: package install commands, option names, and the option Default/Example/Usage blocks all copy the right text and raise the toast, in light and dark themes.


raboof (approved):
neat!

Issue #1270, opened 2026-05-09
feature request: link to options from package


aaaaargZombies:
How feasible would it be to show or link to the options for a package in the "▾▾▾ Show more package details ▾▾▾" section?

It would be very nice to see an overview of all options available for package rather than searching in the options tab and hoping you find the correct ones.

PR #1315, opened 2026-06-05
feat(frontend): link from a package to its NixOS options search

ELHart05:
Addresses #1270.

From a package's expanded details there was no quick way to see which NixOS options relate to it; you had to switch to the Options tab and retype the name. This adds a "NixOS options" line to the package details with a link that runs the options search for the package, carrying the current channel over.

It only shows for nixpkgs packages, not third-party flake packages, which don't have NixOS options.

Tested against the production index: opening firefox shows "Search NixOS options for firefox" linking to /options?channel=unstable&query=firefox, which returns the programs.firefox.* options.

### DIFF
--- a/frontend/src/Page/Packages.elm
+++ b/frontend/src/Page/Packages.elm
@@ -654,6 +654,21 @@ viewResultItem nixosChannels channel showInstallDetails show item =
                         )
                 ]
 
+        nixosOptions =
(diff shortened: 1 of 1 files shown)


raboof (approved):
Cool!

FWIW for flakes perhaps it could link to https://search.nixos.org/flakes?type=options

PR #1317, opened 2026-06-06
feat(frontend): link flake packages to their flake options search

ELHart05:
Follow-up to #1315 (issue #1270). In that review, @raboof noted the package options link could work for flakes too, pointing at the flakes options search.

#1315 only added the "NixOS options" link for nixpkgs packages. This adds the equivalent for flake packages: a "Flake options" link that runs the flakes options search for the package name (`/flakes?type=options&query=...`).

nixpkgs packages are unchanged and still link to `/options` with the channel carried over.

### DIFF
--- a/frontend/src/Page/Packages.elm
+++ b/frontend/src/Page/Packages.elm
@@ -654,20 +654,29 @@ viewResultItem nixosChannels channel showInstallDetails show item =
                         )
                 ]
 
-        nixosOptions =
-            if item.source.flakeUrl == Nothing then
-                div []
-                    [ h4 [] [ text "NixOS options" ]
-                    , p []
-                        [ a [ href ("/options?channel=" ++ channel ++ "&query=" ++ item.source.attr_name) ]
-                            [ text "Search NixOS options for "
-                            , em [] [ text item.source.attr_name ]
+        optionsLink =
+            let
+                searchLink heading label url =
+                    div []
+                        [ h4 [] [ text heading ]
+                        , p []
+                            [ a [ href url ]
+                                [ text label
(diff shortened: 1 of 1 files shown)


KiaraGrouwstra (approved):

KiaraGrouwstra:
this looks like a step forward - tho one failure case here seems `default` flake attributes, which under the current implementation will suggest option searches for `default`

raboof:
> one failure case here seems `default` flake attributes, which under the current implementation will suggest option searches for `default`

good point - maybe in that case we should propose the last segment of the url as search term?

ELHart05:
Seems good? what approach you wanna go with, I can work on it ofc

KiaraGrouwstra:
if i understand correctly that'd be like taking the last word of the `q=` query param as the alternative to `default` in the suggested option search?

ELHart05:
Pushed a commit for this, went with @raboof's idea of using the last segment of the flake url.

So a `github:owner/repo` flake with a `default` package now searches options for `repo` instead of `default`. The repo name already comes off the result, so the link just uses that.

@KiaraGrouwstra on the `query` param route: I left that one out because the last word of the search box isn't really tied to the package you clicked on, and using it would mean threading the query down into each result row. Keying off the flake ref felt more reliable here.

I only special case `default` and fall back to the attr name when there's nothing to derive (an unresolved flake origin, say), so the other attributes stay as they were.

KiaraGrouwstra:
thanks, this approach looks good to me - you as well, @raboof?

raboof (approved):

PR #1346, opened 2026-06-18
remove `_class` mentions` from option results' snippets

### LINKED ISSUE #1344: _class incorrectly shown in option usage example
When viewing the usage example for an option on [https://search.nixos.org/options](https://search.nixos.org/options), `_class = "nixos";` is incorrectly shown as part of it. An example from `programs.bash.enable`'s usage example has been copied below.

```nix
# configuration.nix
{
  _class = "nixos";
  programs.bash.enable = true;
}
```

KiaraGrouwstra:
Closes #1344.

### DIFF
--- a/frontend/src/Page/Options.elm
+++ b/frontend/src/Page/Options.elm
@@ -449,8 +449,7 @@ asHighlightPreCode value =
         ]
 
 
-{-| Render a "Usage" section showing how to use this option in the appropriate
-context, including `_class` to clarify which module system it belongs to.
+{-| Render a "Usage" section showing how to use this option.
 -}
 viewUsageSnippet : ResultItemSource -> List (Html Msg)
 viewUsageSnippet source =
@@ -497,7 +496,6 @@ viewUsageSnippet source =
                 ( Just pkg, Just mod_ ) ->
                     usage
                         ("system.services.<name> = {\n"
-                            ++ "  _class = \"service\";\n"
                             ++ "  imports = [ pkgs."
                             ++ pkg
                             ++ ".services."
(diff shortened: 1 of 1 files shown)


KiaraGrouwstra:
arguably these were more correct than the suggested file names mentioned (`configuration.nix` is now contending for NixOS entrypoint with both `system.nix` as well as with flakes), but yeah, unfortunately `_class` may have been mostly confusing to newer users. perhaps we could find a better place for them some time.

eveeifyeve:
It would be good to make the modular services changed to state the nixos example, so that it's more explicit the example is for nixos. Like adding `# configuration.nix` to the top of the file could do it. With home-manager, darwin, etc people can find out with their respective docs or options defined.

KiaraGrouwstra:
> so that it's more explicit the example is for nixos

hm. so.
is that technically correct right now tho?
cuz you would get the generic functionality (which is worth whatever your system shims do with that), while currently at least one service (php-fpm) out of the like 4 seems to have finix bits there as well.

now, the existing services there may have some escape hatches for nixos, but at that point we've maybe reached a semantic discussion there.

the underlying semantic question right now then becomes: is a modular service something that works on your system, or is it a service implementing the generic interface, rendering 'it works on system x' technically outside of the responsibility of what a modular service is, at least until the interface (and the system's shims, without requiring the escape hatches) make it work?

i think the proper solution there would be explicitly segregating the generic bits vs the system-specific escape hatches so one would know what they'd get, but that isn't where we're at today.

raboof (approved):
I agree the `_class` is currently confusing, let's merge this to remove it and make further improvements in follow-up PRs

PR #1357, opened 2026-06-20
mention matrix channel in readme

KiaraGrouwstra:


raboof (approved):
it's a bit uncomfortable since it's not part of the nixos space and AFAICS because there's no moderators left it can't be made part of that space anymore either, but good to make it more discoverable I suppose.

PR #1368, opened 2026-06-30
fix(ci): better deal with index import errors to never silently serve a stranded ES import index

KiaraGrouwstra:
As reported by @kranzes, our search has currently failed to index some recent options like [`services.nginx.lua`](https://search.nixos.org/options?channel=unstable&query=services.nginx.lua) and [`system.boot.extraInitrd`](https://search.nixos.org/options?channel=unstable&query=system.boot.extraInitrd), despite both of these being part of the unstable branch already.

This turned out to be because a transient `504` mid-push could strand a half-built nixos-unstable index: the alias was never flipped, and every subsequent 2-hourly run was a silent green no-op because the `abort` strategy treated "index exists" as success.

This change includes a few fixes to deal with this:

- Retry transient `5xx`/`429`/transport errors on bulk push with bounded exponential backoff (`2s`/`4s`/`8s`/`16s`, max 5 attempts). Permanent `4xx` and per-item failures are not retried. Add `is_retryable_status` (unit-tested).
- Clear-on-failure: if push or alias-write fails in the create path (`Abort`/`Recreate`), best-effort delete the just-created index so the next run rebuilds rather than treating the partial index as "already exists". Never clears under `Ignore` (append).
- Abort guard: on `IndexExistsError`, check whether the alias already points at this index via `alias_points_at`. Current -> `Ok`; stranded -> error so the job fails and the index is rebuilt.
- Warmup tripwire: assert a doc-count floor on the alias target, and assert the alias resolves to the index this run just pushed (emitted as `FLAKE_INFO_PUSHED_INDEX`). A non-zero exit fires the existing `if: failure()` issue creation.

Also drop the `dbg!(response)` macros that spammed CI stderr.

### DIFF
--- a/.github/workflows/import-to-elasticsearch.yml
+++ b/.github/workflows/import-to-elasticsearch.yml
@@ -66,53 +66,16 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         shell: sh
         run: |
-          # Capture stdout so we can read the FLAKE_INFO_PUSHED_INDEX line that
-          # flake-info emits, and assert the alias resolves to it in Warmup.
-          # `set -e` (sh default off) plus an explicit status check ensures a
-          # failed import still fails the step despite the `tee` pipe.
-          set -e
-          nix run --accept-flake-config .#flake-info -- --push --elastic-schema-version=$(nix eval --raw --file ./version.nix import) nixpkgs ${{ matrix.channel }} > flake-info.out
-          cat flake-info.out
-
-          PUSHED_INDEX="$(sed -n 's/^FLAKE_INFO_PUSHED_INDEX=//p' flake-info.out | tail -n1)"
-          echo "pushed index: $PUSHED_INDEX"
-          echo "PUSHED_INDEX=$PUSHED_INDEX" >> "$GITHUB_ENV"
+          nix run --accept-flake-config .#flake-info -- --push --elastic-schema-version=$(nix eval --raw --file ./version.nix import) nixpkgs ${{ matrix.channel }}
 
       - name: Warmup ${{ matrix.channel }} channel
         if: github.repository_owner == 'NixOS'
         shell: sh
         run: |
-          INDEX="latest-$(nix eval --raw --file ./version.nix import)-nixos-${{ matrix.channel }}"
-
           for i in 1 2 3
           do
-            curl -sS "${{ secrets.ELASTICSEARCH_URL2 }}/$INDEX/_search" | jq -ce '.took // error'
+            curl -sS ${{ secrets.ELASTICSEARCH_URL2 }}/latest-$(nix eval --raw --file ./version.nix import)-nixos-${{ matrix.channel }}/_search | jq -ce '.took // error'
           done
-
-          # Doc-count floor: a half-built index that somehow got aliased would
-          # have far fewer docs than a real import. Fail the job if the alias
-          # target is suspiciously small so the import is rebuilt rather than
-          # silently serving stale/partial data.
-          MIN_DOCS=100000
-          COUNT="$(curl -sS "${{ secrets.ELASTICSEARCH_URL2 }}/$INDEX/_count" | jq -e '.count')"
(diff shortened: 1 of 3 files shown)


.github/workflows/import-to-elasticsearch.yml
```diff
@@ -66,17 +66,54 @@ jobs:
...
+          # have far fewer docs than a real import. Fail the job if the alias
+          # target is suspiciously small so the import is rebuilt rather than
+          # silently serving stale/partial data.
+          MIN_DOCS=100000
+          COUNT="$(curl -sS "${{ secrets.ELASTICSEARCH_URL2 }}/$INDEX/_count" | jq -e '.count')"
+          echo "Index $INDEX has $COUNT docs (floor $MIN_DOCS)"
+          if [ "$COUNT" -lt "$MIN_DOCS" ]; then
+            echo "Doc count $COUNT below floor $MIN_DOCS; alias target is stale or half-built" >&2
+            exit 1
+          fi
```
raboof: This never happens/happened, right? I'm not sure if this is worth it.

.github/workflows/import-to-elasticsearch.yml
```diff
@@ -66,17 +66,54 @@ jobs:
...
+          # run's index is still aliased and a no-op run would otherwise look
+          # green. Skipped only if no index name was captured (older binary).
+          if [ -n "$PUSHED_INDEX" ]; then
+            if ! curl -sS "${{ secrets.ELASTICSEARCH_URL2 }}/$INDEX/_alias" | jq -e --arg idx "$PUSHED_INDEX" 'has($idx)' > /dev/null; then
+              echo "Alias $INDEX does not resolve to freshly pushed index $PUSHED_INDEX" >&2
+              curl -sS "${{ secrets.ELASTICSEARCH_URL2 }}/$INDEX/_alias" | jq -r 'keys[]' >&2 || true
+              exit 1
+            fi
+            echo "Alias $INDEX correctly resolves to $PUSHED_INDEX"
+          fi
```
raboof: Isn't this already covered by the new `Index {index} exists but alias {a} does not point at it` logic on the rust side?

flake-info/src/bin/flake-info.rs
```diff
@@ -463,29 +463,71 @@ async fn push_to_elastic(
...
     let ensure = es.ensure_index(&config).await;
     if let Err(ElasticsearchError::IndexExistsError(_)) = ensure {
-        // abort on abort
-        return Ok(());
+        // The index already exists under the Abort strategy. This is only a
+        // legitimate no-op if the alias already points at it (a previous run
+        // completed successfully). If the alias does not point here, the index
+        // is stale/half-built (e.g. a prior push died before flipping the
+        // alias) - fail so the job is red and the index gets rebuilt, rather
+        // than silently returning green forever.
```
raboof: how does the index get rebuilt in this scenario?

flake-info/src/elastic.rs
```diff
@@ -228,6 +226,13 @@ lazy_static! {
     });
 }
 
+/// Whether an HTTP status code from a bulk push is a transient failure worth
+/// retrying. Covers ES/proxy overload (429) and gateway/upstream errors
+/// (502/503/504). All other statuses are treated as deterministic.
+pub fn is_retryable_status(code: u16) -> bool {
+    matches!(code, 429 | 502 | 503 | 504)
```
raboof: are you sure these are all retryable, e.g. that they don't leave the index in a partially-updated state where performing the request again would create duplicates? I guess I'm fine with just trying it out.

raboof (commented):
Overall great! Some comments/questions, and seems to need a format.

flake-info/src/bin/flake-info.rs
KiaraGrouwstra: thanks, i fixed it to now clear it so the next job trigger should go thru at least

flake-info/src/elastic.rs
KiaraGrouwstra: fair - dropped them for now

raboof (approved):

PR #1370, opened 2026-07-01
frontend: enhance logo

Malix-Labs:
Style change should not be in CI


Malix-Labs:
CC @Eveeifyeve, since you're the last person in https://github.com/NixOS/nixos-search/commits/main/.github/workflows/switch-logo.yml

KiaraGrouwstra (approved):

eveeifyeve (approved):
I think this is good and this does it for the favicon as well, right?

Malix-Labs:
> this does it for the favicon as well, right?

Not that I know of, it wasn't the case before either

I can add that if needed

Malix-Labs:
Also

https://github.com/NixOS/nixos-search/blob/main/frontend/src/assets/images/nix-logo.png

and

https://github.com/NixOS/nixos-search/blob/main/frontend/src/assets/images/nix-logo-pride.png

are not at the same scale

And they could be SVGs, taken from https://github.com/NixOS/branding

Malix-Labs:
I also did make it use https://github.com/NixOS/branding !

KiaraGrouwstra (approved):

Sigmanificient (approved):
Pride month is already over :(

PR #1375, opened 2026-07-08
feat(frontend): port improved ES ranking signals to search

KiaraGrouwstra:
Add three ranking signals from @Mic92's [query enhancements](https://github.com/NixOS/nixos-search/pull/1262#issuecomment-4912257082):

- Exact `term` + `prefix` boost on the primary keyword field (`package_attr_name` / `option_name`) so canonical attrs like `gcc`, `docker`, `nodejs_26` float above nested variants.
- A `constant_score` phrase bonus for multi-word queries so results matching the query as a phrase in the description rank higher.
- A packages-only top-level `rescore` `function_score` script keyed on attr-name length so `nodejs_26` beats `nodejs-slim_26`.

### DIFF
--- a/frontend/src/Page/Options.elm
+++ b/frontend/src/Page/Options.elm
@@ -836,6 +836,8 @@ makeRequestBody types query from size sort =
         , ( "service_package", 3.0 )
         , ( "service_packages", 3.0 )
         ]
+        [ "option_description^3" ]
+        Nothing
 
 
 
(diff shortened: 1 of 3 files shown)


raboof (approved):

Jappie3:
image

any search on search.nixos.org/packages now returns 400, possibly related to this PR?

edit: seems like Kiara is aware already :sweat_smile:

PR #1383, opened 2026-07-12
comment fenix, whose x86_64-darwin now throws

### LINKED ISSUE #1382: "i3GapsSupport" shifting to "i3Support" in home-manager option "services.polybar.package"
the arg `i3GapsSupport` in "pkgs.polybar.override", the config option of "services.polybar.package", had been renamed to `i3Support`, which has not been updated in nixos-search

i guess it might be due to the merging of i3-gaps, i can't find something called "i3-gaps" in nixpkgs now

KiaraGrouwstra:
Closes #1382.

a better solution would be to not break the periodic import as soon as any flake build fails.

### DIFF
--- a/flakes/manual.toml
+++ b/flakes/manual.toml
@@ -337,10 +337,11 @@ type = "github"
 owner = "nix-community"
 repo = "autofirma-nix"
 
-[[sources]]
-type = "github"
-owner = "nix-community"
-repo = "fenix"
+# x86_64-darwin throws deprecated
+# [[sources]]
+# type = "github"
+# owner = "nix-community"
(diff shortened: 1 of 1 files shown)


raboof:
> a better solution would be to not break the periodic import as soon as any flake build fails.

Jep. The main challenge is to make sure such failures don't end up going unnoticed.

raboof (approved):

Malix-Labs:
> The main challenge is to make sure such failures don't end up going unnoticed.

See #1027 _(for reference, I know you're aware already (I want to fix that so bad but UGH))_

PR #1386, opened 2026-07-13
chore: remove <nixpkgs> impurity and use pure flake overrides

Malix-Labs:
This PR cleans up the option evaluator to remove impurities and reduce duplicate boilerplate:

- **Impurities**: Replaces legacy `<nixpkgs>` paths and `-I` search path overrides with pure `builtins.getFlake "nixpkgs"` and `--override-flake` CLI arguments.
- **DRY Refactor**: Consolidates repetitive `nix eval` Command builder boilerplate in Rust under a new `nix_eval_command` helper, and DRYs up option module evaluation in `flake_info.nix` with a generic `evalOptionsWith` helper.

### DIFF
--- a/flake-info/assets/commands/flake_info.nix
+++ b/flake-info/assets/commands/flake_info.nix
@@ -5,8 +5,9 @@
 let
   resolved = builtins.getFlake input-flake;
 
-  nixpkgs = (import <nixpkgs> { });
-  lib = nixpkgs.lib;
+  nixpkgsFlake = builtins.getFlake "nixpkgs";
+  inherit (nixpkgsFlake) lib;
+  nixpkgs = nixpkgsFlake.legacyPackages.${referenceSystem};
 
   # filter = lib.filterAttrs (key: _ : key == "apps" || key == "packages");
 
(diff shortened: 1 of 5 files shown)


flake-info/assets/commands/flake_info.nix
```diff
@@ -517,19 +531,12 @@ rec {
   packages = lib.attrValues (collectSystems allPackageSets.packages packages');
   apps = lib.attrValues (collectSystems { } apps'); # apps don't need fallback evaluation
   options = readFlakeOptions;
-  home-manager-options =
-    let
-      # Require both `modules/modules.nix` and `modules/lib/stdlib-extended.nix`
-      # to avoid false positives. Other flakes (e.g. `nix-bitcoin`) ship a
-      # `modules/modules.nix` that is unrelated to home-manager; only
-      # home-manager itself also provides the `stdlib-extended.nix` helper
-      # that `readHomeManagerOptions` imports.
```
KiaraGrouwstra, line 526: should we keep the comment still?

KiaraGrouwstra (approved):

flake-info/assets/commands/flake_info.nix
Malix-Labs: I think yes, since the check requiring both files looks arbitrary without it; the design decision is not that obvious, and maybe it would prevent removing it by accident, i'll add it back

Malix-Labs: fixed it in 62370ccfbb6df9be205730bf57c0ab358609d68e

PR #1387, opened 2026-07-13
feat: add nix-darwin options search

### LINKED ISSUE #1384: feat: nix-darwin options
https://github.com/nix-darwin/nix-darwin

## Related

- [ ] #1385
- [x] #1281

Malix-Labs:
Adds nix-darwin options search capability to the indexer and frontend.

CLOSES: #1384

### DIFF
--- a/flakes/manual.toml
+++ b/flakes/manual.toml
@@ -354,6 +354,11 @@ repo = "home-manager"
 
 [[sources]]
 type = "github"
+owner = "nix-darwin"
+repo = "nix-darwin"
+
+[[sources]]
+type = "github"
 owner = "nix-community"
 repo = "nixvim"
 
--- a/flake-info/assets/commands/flake_info.nix
+++ b/flake-info/assets/commands/flake_info.nix
@@ -345,6 +345,22 @@ let
       };
     };
 
+  # Extract options from nix-darwin's module system.
+  # Evaluated separately during the nixpkgs channel import (via
+  # `--override-flake input-flake github:nix-darwin/nix-darwin`) so that
+  # nix-darwin options land in the channel index alongside NixOS options.
+  readDarwinOptions =
+    let
+      darwinModulesPath = "${resolved}/modules/module-list.nix";
+      darwinModuleList = import darwinModulesPath;
+    in
+    evalOptionsWith {
+      modules = darwinModuleList;
+      extraAttrs = {
+        entry_type = "darwin-option";
+      };
+    };
+
   read = reader: set: lib.flatten (lib.attrValues (withSystem reader set));
 
   # Get all package sets by system for potential fallback evaluation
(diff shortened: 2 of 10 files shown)


KiaraGrouwstra:
i think this would need to be split such that a first PR would add it to the back and bump the back's version, then a follow-up would add the front and bump the front's version (and similar for the nixbsd PR)

Malix-Labs:
I wasn't aware of nixbsd, I will check that one!

The other PR is about nix-darwin :)

And are you saying that because otherwise nixos search would be broken down for an hour or so otherwise?
I think we can trigger an elastic run GitHub action so that it wouldn't be needed, no?
It was the same for the licencing shortening PR that I authored and that was merged I think

KiaraGrouwstra:
interesting, i see.
about the versions: separating it would make it explicit that it shouldn't try to use new data before it actually gets dumped. @raboof may know more on this.
it may be that if we keep it in a single PR then that'd leave this new bit broken for just a few hours until the new index lands, yeah.

raboof:
> And are you saying that because otherwise nixos search would be broken down for an hour or so otherwise?
> I think we can trigger an elastic run GitHub action so that it wouldn't be needed, no?

If the change to the schema is backwards-incompatible we should indeed split the change in two parts: one changing the export and index version used by the export, and then once the new indices are populated, one changing the frontend and index version used by the frontend.

> It was the same for the licencing shortening PR that I authored and that was merged I think

Indeed it looks like we merged #1329 as one part, and without changing the schema version. I'm fairly sure that means the change was backwards-compatible enough that the new frontend worked reasonably well against the old schema - otherwise we would've heard :) .

> it may be that if we keep it in a single PR then that'd leave this new bit broken for just a few hours until the new index lands, yeah.

If it's just the new feature that's broken that might be fine, but if the whole search is broken then even a few hours is quite painful (and it's hard to guarantee it'll be just a few hours).

Malix-Labs:
> that means the change was backwards-compatible enough

Yep, it indeed was

I think that one also qualifies, except maybe https://github.com/NixOS/nixos-search/pull/1387/changes#diff-00f50416e7bf1d3c0e6dfb802a21816d00ab1cc3fd5c9cecca075a3df8a28756R57 (R57) ?

The link is broken because GitHub and Ruby on Rails are perfectly fine and qualitative software, so here is a screenshot:

image

I'll test it on localhost

Malix-Labs:
Yep, the change is also completely backwards compatible, Elm doesn't crash on any missing variable, everything works (except the darwin tab will shows nothing until elastic catches up of course)

You can also verify this locally on your machine (`npm run dev` on `/frontend`, then open `http://localhost:3000/`)

image

KiaraGrouwstra (approved):

flakes/manual.toml
```diff
@@ -352,6 +352,11 @@ type = "github"
 owner = "nix-community"
 repo = "home-manager"
 
+[[sources]]
+type = "github"
+owner = "nix-darwin"
+repo = "nix-darwin"
+
```
Malix-Labs, line 359: Am I dumb or what


```suggestion
```

PR #1379, opened 2026-07-10
Introduce the new astro frontend rewrite.

eveeifyeve:
Not working/ready yet, but just getting the idea out there.

> NOTE: Disclaimer, this is done under paid work and this pr is also AI assisted.

PR #1395, opened 2026-07-14
fix(frontend): remove colliding native autocomplete

Sigmanificient:
## before

image

## after

image

### DIFF
--- a/frontend/src/Search.elm
+++ b/frontend/src/Search.elm
@@ -1234,6 +1234,7 @@ viewSearchInput nixosChannels outMsg categoryName selectedChannel searchQuery ma
                 [ input
                     [ type_ "text"
                     , id "search-query-input"
+
                     -- not really sure how to make this better, sadly
                     -- TODO: improve me
                     , autocomplete (categoryName == "3rd-party flake packages")
--- a/frontend/src/Search.elm
+++ b/frontend/src/Search.elm
@@ -63,6 +63,7 @@ import Html
 import Html.Attributes
     exposing
         ( attribute
+        , autocomplete
         , autofocus
         , class
         , classList
@@ -1233,6 +1234,9 @@ viewSearchInput nixosChannels outMsg categoryName selectedChannel searchQuery ma
                 [ input
                     [ type_ "text"
                     , id "search-query-input"
+                    -- not really sure how to make this better, sadly
+                    -- TODO: improve me
+                    , autocomplete (categoryName == "3rd-party flake packages")
                     , autofocus True
                     , placeholder <| "Search for " ++ categoryName
                     , onInput (outMsg << QueryInput)
--- a/frontend/src/Search/Typeahead.elm
+++ b/frontend/src/Search/Typeahead.elm
@@ -33,7 +33,7 @@ submit-on-enter UX.
 import Base64
 import Dict exposing (Dict)
 import Html exposing (Html, a, li, span, text, ul)
-import Html.Attributes exposing (class, href)
+import Html.Attributes exposing (class, classList, href)
 import Http
 import Json.Decode
 import Json.Decode.Pipeline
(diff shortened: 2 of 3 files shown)


Sigmanificient:
@Malix-Labs You think you could help me with that one?

Malix-Labs:
@Sigmanificient you want a review?

Sigmanificient:
i'd like help for the accessibility part :D

KiaraGrouwstra (approved):

Malix-Labs:
Whoops I just logged in for the day and did go to sleep after https://github.com/NixOS/nixos-search/pull/1395#issuecomment-4974575411, and it's already merged @Sigmanificient

raboof:
;) I figured the change as-is was already a nice improvement - further improvements still welcome of course though!

Malix-Labs:
There are still problems, for example, try to go from

https://search.nixos.org/packages?channel=26.05&query=eh (the autocomplete doesn't show up)
and then click on the "options" navigation button (the autocomplete also doesn't show up)

frontend/src/Search.elm
```diff
@@ -1233,6 +1234,10 @@ viewSearchInput nixosChannels outMsg categoryName selectedChannel searchQuery ma
                 [ input
                     [ type_ "text"
                     , id "search-query-input"
+
+                    -- not really sure how to make this better, sadly
+                    -- TODO: improve me
```
Malix-Labs, line 1239: What do you mean by "better" ?
Do you mean "make the non-native autocomplete work with flakes instead of fallbacking to the native one for them" ?

frontend/src/Search/Typeahead.elm
```diff
@@ -355,12 +355,13 @@ update options nixosChannels searchType activeSource channel query msg model =
...
 viewDropdown model =
-    if not model.preferStatic || not model.visible || List.isEmpty model.suggestions then
-        text ""
+    ul
+        [ class "typeahead-suggestions"
 
-    else
-        ul [ class "typeahead-suggestions" ]
-            (List.map viewSuggestion model.suggestions)
+        -- TODO: make it work with focus-within for accessibility!
```
Malix-Labs, line 361: Was there any particular difficulty with adding the `:focus-within` CSS pseudo class directly?

Malix-Labs:
Also, I am a native enjoyer and would have gotten rid of the non-native autocomplete instead of getting rid of the native one

Malix-Labs:
Actually I want to PR the idea in https://github.com/NixOS/nixos-search/pull/1395#issuecomment-4982013530

Since it is possible to populate a datalist dynamically

Malix-Labs:
An alternative is to get rid of autocomplete altogether and make the result reload oninput

Since that this product focus on search, I always find this UX largely superior

Or even better (server-load wise) would be debounced live results with no separate autocomplete at all
+ only show the first 10 results + infinite scrolling + streaming results

I will try to see what I can do about that

Or maybe keep that in #1379 ?

frontend/src/Search.elm
Sigmanificient: The check is hard coded as `categoryName == "3rd-party flake packages"`, not great imho bc it relies on a representation string!

Sigmanificient:
> *Also, I am a native enjoyer and would have gotten rid of the non-native autocomplete instead of getting rid of the native one*

@Malix-Labs Yes there maybe something to do with `<select>` based autocompletion?

frontend/src/Search.elm
Eveeifyeve: Eventually with  #1379 I want this to be generalized, So autocomplete just happens when you are searching, Kinda of like google when you searching something, Sept the autocomplete will show packages, The root set of module options (which includes modular services), functions & documentation.


Root set of module options I mean stuff like it contains like `virutalization.docker.*` will just become `virutalization.docker` in autocomplete. Until you have specified exactly `virutalization.docker` then it will become the other options.

PR #1406, opened 2026-07-16
frontend: replace custom sort options dropdown with native select element

Malix-Labs:
CC @Sigmanificient

## Before

image

image

## After

Still keeping the "clicking on the whole label does open the dropdown" behavior, which is a good UX I wanted to keep

image


KiaraGrouwstra (approved):

Malix-Labs:
2dc85ed109c11ed614d26b150580bdc4ea8687f3 makes it match the branding better:

image

It is still not exactly the same as before, but at least it removes the previous hacks and bootstrap garbage

Malix-Labs:
Okay now I nailed it

image

It looks even better than before imo

PR #1409, opened 2026-07-16
frontend: general enhancements

Malix-Labs:


frontend/src/scss/index.scss
```diff
@@ -138,7 +141,7 @@
     @include outline-visible;
   }
 
-  button.btn.btn {
+  .btn.btn {
```
KiaraGrouwstra, line 144: hm, prob an improvement but maybe redundant still?

KiaraGrouwstra (approved):

frontend/src/scss/index.scss
Malix-Labs: Well yes very probably, there is so much to cleanup, the whole scss taken from bootstrap is a dumpster fire

Malix-Labs:
Btw, the search input elements are still not well architectured enough, so there are weird edge cases with responsiveness and such. you can see that easily in 3rd party flakes. I'll fix this

Malix-Labs:
@KiaraGrouwstra it is A NIGHTMARE to fix because it is starting to get close to the level of indirection that I clowned in [the tweet](https://x.com/Malix_Labs/status/2077521369663742252?s=20)

I tried grids, and every time I fix an issue, something else gets side-affected by it.
It's quite funny that Elm was chosen (given that it is chosen precisely to reduce JS side-effects) while turning the CSS into a side-effect spaghetti SCSS

I even tried with more powerful models, but the token burning is massive since there isn't really any way to mentally infer that many amount of indirection

I'm starting to understand the pain of my Nix untangling the perl scripts

KiaraGrouwstra:
makes sense, yeah.
hopefully ditching elm/scss sooner or later might help alleviate this.

Malix-Labs:
I though Elm was the main problem (it is still a problem, I would prefer something reactive and performant like signals such as in Svelte and Solid), but the main problem is clearly scss, sigma was right about it

KiaraGrouwstra:
fair enough - thanks to @Sigmanificient then as well!

Issue #1421, opened 2026-07-18
ranking quality benchmark


KiaraGrouwstra:
one recent topic here has been the quality of query result rankings, raised at #1418 as well as at https://github.com/NixOS/nixos-search/pull/1262#issuecomment-4912257082 (later implemented in #1375).

movement on this simultaneously raises the question if we could measure such ranking quality, on the one hand to get a sense of how we're doing there, while on the other hand to prevent potential regressions on this front.

it looks like @hsjobeki's https://hsjobeki.github.io/nixos-search-queries/ might make for an interesting starting point on this, geared toward our dataset, while identifying various query groups of interest, as well as specifying some initial ranking expectations on these, to make things measurable.

KiaraGrouwstra:
closed in #1425

PR #1422, opened 2026-07-18
frontent/search: add fuzzy matching to elastic parameters

hsjobeki:
This enables fuzzy matching capabilities of elastic search.

Which in my search performance evaluation showed to add better support for `typo` and `prefix` queries.

https://github.com/hsjobeki/nixos-search-queries

### DIFF
--- a/frontend/src/Search/Query.elm
+++ b/frontend/src/Search/Query.elm
@@ -297,8 +297,25 @@ searchFields positiveWords mainFields fields =
                     ]
               )
             ]
+
+        fuzzyMatch : List ( String, Json.Encode.Value )
+        fuzzyMatch =
+            [ ( "multi_match"
+              , Json.Encode.object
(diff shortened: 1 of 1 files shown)


KiaraGrouwstra:
- the used relative `weight` sounds arbitrary, so we might need to figure out what makes sense here
- this new sub-clause does multiple things, raising the question if any of them should perhaps also be reused in (the) other sub-clauses?
- in order to evaluate this change, we might want some benchmark (#1421) to compare it to our status quo, given that in the presence of our other sub-queries, results on the original benchmark (performed on an isolated query) might be affected.

KiaraGrouwstra:
this may need to come after #1428

KiaraGrouwstra (approved):

KiaraGrouwstra:
with #1442 down we could now retry something like this

PR #1425, opened 2026-07-20
frontend/search: add benchmark ci report

hsjobeki:
Adds a benchmark report for search results

Inspired by https://github.com/hsjobeki/nixos-search-queries.

- Added 150 curated queries in different categories.
- Report metrics for each category.

New workflow should post a comment on PRs touching these files.

TODOs for follow up PRs:

- Before / After comparison
- Also move "attr_name" boost and other query modifications into one file.
- Test backend updates

Assisted-by: Opus-4-8

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,55 @@
+name: "Frontend: Relevance benchmark"
+
+on:
+  pull_request:
+    paths:
+      - "frontend/src/Search.elm"
+      - "frontend/src/Page/Packages.elm"
+      - "frontend/src/Page/Options.elm"
+      - "frontend/src/Benchmark.elm"
+      - "frontend/benchmark/**"
+      - ".github/workflows/benchmark.yml"
+
+# This job builds and runs untrusted PR code
+# We produce a markdown report as artifact; which the trusted workflow downloads
+permissions:
+  contents: read
+
+jobs:
+  benchmark:
+    name: "Relevance benchmark"
+    runs-on: ubuntu-latest
+    continue-on-error: true
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7.0.0
+
+      - name: Set up Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: "24"
+          cache: "npm"
+          cache-dependency-path: frontend/package-lock.json
+
+      - name: Install dependencies
+        run: npm ci
+        working-directory: frontend
+
+      - name: Run relevance benchmark
+        run: npm --prefix frontend --silent run benchmark > benchmark_report.md
+
+      - name: report
+        run: |
+          echo "UPLOADING BENCHMARK REPORT"
+          cat benchmark_report.md
+          echo "${{ github.event.pull_request.number }}" > pr_number.txt
+
+      - name: Upload report artifact
+        uses: actions/upload-artifact@v4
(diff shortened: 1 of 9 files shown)


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,47 @@
...
+        run: npm ci
+        working-directory: frontend
+
+      - name: Run relevance benchmark
+        run: |
+          SCHEMA=$(grep -oP 'frontend = "\K[0-9]+' version.nix)
+          npm --prefix frontend run benchmark -- --schema "$SCHEMA" > benchmark_report.md
+
+      - name: Post report as PR comment
+        uses: marocchino/sticky-pull-request-comment@v2
```
KiaraGrouwstra: i guess we'll find out if this triggers on re-push / force push as soon as you push again to fix the formatting

frontend/benchmark/run.mjs
```diff
@@ -0,0 +1,216 @@
...
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const FRONTEND_DIR = resolve(__dirname, "..");
+
+const { values: args } = parseArgs({
+    args: process.argv.slice(2),
+    options: {
+        queries: { type: "string", default: join(__dirname, "queries.json") },
+        channel: { type: "string", default: "nixos-unstable" },
+        schema: { type: "string", default: "48" },
```
KiaraGrouwstra: maybe note in the PR description that we should either try and make this dynamic or at least document this is another place where we should update it (when updating the back-end i guess?)

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,47 @@
...
+
+# Runs when the client-side Elasticsearch query changes
+# Reports current query performance statistics
+
+on:
+  pull_request:
+    paths:
+      - "frontend/src/Search.elm"
+      - "frontend/src/Page/Packages.elm"
+      - "frontend/src/Page/Options.elm"
```
KiaraGrouwstra, line 8: maybe it'd be nicer if this were better factored out so we wouldn't need to rerun the benchmark each time something changes in the front-end

hsjobeki: Deliberately left this for the next PR to keep things reviewable

KiaraGrouwstra:
so the action log now shows an upload of this zip file containing the PR number (in a file) and the actual benchmark report.
if we could figure out posting the latter in a PR comment (in a least-privilege kinda fashion), that'd be cool still.

hsjobeki:
Yes this is intentional. So the untrusted code in the PR produces and artifact which contains the report. 
Once our own trusted workflow runs it downloads the artifact and posts a comment

KiaraGrouwstra:
CI:

```
[benchmark] scoring 150 queries against latest-48-nixos-unstable
node:internal/modules/run_main:107
    triggerUncaughtException(
    ^

[TypeError: fetch failed] {
  [cause]: Error: read ETIMEDOUT
      at TLSWrap.onStreamRead (node:internal/stream_base_commons:216:20) {
    errno: -110,
    code: 'ETIMEDOUT',
    syscall: 'read'
  }
}
[1](https://github.com/NixOS/nixos-search/actions/runs/29763093600/job/88422344077?pr=1425#step:5:19)
```

KiaraGrouwstra (approved):

PR #1437, opened 2026-07-21
flake-info: add package metadata from repology 

hsjobeki:
This allows us to use:

- popularity
- reverse dependency count

For improving search ranking. Note that this requires a second PR to use the ranking once the index is updated.

Used @Mic92 's work and added:

- Retry on failed fetch with exponential back-off. (Network can be flakey; and we fetch information about all repology packages, paginated)
- Run the job once per day; instead of every 2 hrs. This seems still too much; but if it ever breaks, we can name the day.

Assisted-by: opus-4-8

### DIFF
--- a/.github/workflows/update-repology-counts.yml
+++ b/.github/workflows/update-repology-counts.yml
@@ -1,0 +1,63 @@
+name: "Daily Repology counts refresh"
+
+on:
+  workflow_dispatch:
+  schedule:
+    # Daily, offset from the 2-hourly imports. Repology refreshes its
+    # `nix_unstable` data roughly daily, so this is a generous ceiling.
+    - cron: "0 3 * * *"
+
+permissions:
+  contents: read
+  issues: write
+
+jobs:
+  update-repology-counts:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checking out the repository
+        uses: actions/checkout@v7.0.0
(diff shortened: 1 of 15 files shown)


.github/workflows/update-repology-counts.yml
```diff
@@ -0,0 +1,63 @@
+name: "Daily Repology counts refresh"
+
+on:
+  workflow_dispatch:
+  schedule:
+    # Daily, offset from the 2-hourly imports. Repology refreshes its
```
KiaraGrouwstra, line 6: daily might be a lot since users may notice the changes in signal less so than repology would notice the daily scrapes?

.github/workflows/update-repology-counts.yml
```diff
@@ -0,0 +1,63 @@
...
+    # `nix_unstable` data roughly daily, so this is a generous ceiling.
```
KiaraGrouwstra, line 7: do we even need them for nixpkgs info?

.github/workflows/update-repology-counts.yml
hsjobeki: If the workflow ever breaks you get a specific day where it did. 
But I think once per week would still suffice for data freshness

KiaraGrouwstra (approved):

Issue #1438, opened 2026-07-21
Search Quality Regression


lveneris:
## Brief

I noticed earlier when I was searching for `continuwuity` that I got a plethora of completely unrelated results, mostly related to continuous integration engines. While it was quite amusing to discover that the actual package I was looking for, `matrix-continuwuity`, was on the last page, I do not think this regression was intentional.

Thank you for your work on this project.

## References

It may have been introduced by this commit, identified by @drainpixie: https://github.com/NixOS/nixos-search/commit/3b7b5cce56df0f5dda810424d3116fca5bc05729.

For demonstration:

<img width="1220" height="967" alt="Image showing the search performance of search.nixos.org." src="https://github.com/user-attachments/assets/44e0b94b-0c09-442a-b20f-a50fb35623d9" />

<img width="1220" height="967" alt="Image showing the search performance of search.nixos.org" src="https://github.com/user-attachments/assets/1f508721-c275-43f5-9782-65048c8e441c" />

KiaraGrouwstra:
thanks for filing!

could you elaborate what on the `not-three`? is that an existing package?

lveneris:
and thank you for getting back to me! I use nixos-search to check if a package exists or not, sometimes, which normally gives me a fairly quick answer. on this particular occasion, it turns out [`not-three`](https://github.com/not-three/main) is not packaged (I had to check the repo myself to confirm), but there still should not be so many completely unrelated packages in the results.

raboof:
@hsjobeki does your benchmark (#1425) confirm this regression or should we improve it? Is there anything we can do to improve the ranking again or should we roll back?

hsjobeki:
I think this single query "regression" will probably be improved once we use the repology metadata for ranking as well. 

Also consider the countersied before calling it a regression: people who got improved results usually do not file issues. We have improved the situation for statistically speaking more people than it got worse for some edge cases.

raboof:
Comparing the benchmark of #1422  against the earlier #1425 :

### Overall

| metric | value before | value after |
| --- | --- | --- |
| Success@10 | 0.580 | 0.740 |
| MRR | 0.442 | 0.562 |
| Recall@10 | 0.483 | 0.590 |

### By category

| category | n | Success@10 before | MRR before | Success@10 after | MRR after |
| --- | --- | --- | --- | --- | --- |
| dotted | 25 | 0.840 | 0.480 | 0.480 |
[...]

PR #1441, opened 2026-07-22
add `continuwuity` as benchmark query

KiaraGrouwstra:
added as per report at #1438.


raboof:
> g151 	continuwuity 	exact 	0 	0.000 	opt:services.matrix-continuwuity.enable, opt:services.matrix-continuwuity.settings.global.database_path, opt:services.matrix-continuwuity.settings.global.port

Should we split the benchmarks between package search and option search? These are fine results for the options search, but it's the package search that's not performing well.

raboof (approved):
seems like an improvement in any case

Issue #1434, opened 2026-07-21
Move flake-info eval to nix implementation.


eveeifyeve:
#### The Problem: 
`flake-info` reimplements functionality in Rust that Nix already provides natively, which adds unnecessary overhead and complexity to `nixos-search`'s CI. It also doesn't account for future improvements and changes in Nix, such as [Flake Schemas](https://github.com/NixOS/nix/pull/8892) and adds overhead into expanding support for flake options, packages, functions, and documentation. So for the reasons this becomes overly complex under the current architecture compared to an Nix-based approach, it also make it's easier to deprecate the old special types such as `home-manager-option` & `darwin-option` with just the type being `option` and a column for the flake_ref attribute set. 

#### Solution: 
Instead of relying on custom Rust logic to inspect flakes, we transition to a native Nix implementation. There is a working prototype has already been created by @Malix-Labs here: https://gist.github.com/Malix-Labs/b33fbec809d7a9b5334747a22bfd61da, which this approach allows us to leverage Nix directly for extracting metadata, making future additions (e.g., options, docs, functions) from flakes making it significantly easier to maintain.

### What is there todo:
- [ ] Introduce @Malix-Labs' prototype into the codebase and introduce a new indices that is aliased under something like: `latest-*-nixos-*-beta`.
- [ ] Introduce it to the new Astro Frontend Rewrite: https://github.com/NixOS/nixos-search/pull/1379 (@Eveeifyeve)
- [ ] Start migrating to the prototype into a production-ready implementation within `nixos-search` and deprecate/remove redundant `flake-info` components.

---

* Related links: [Redesign pr diagram](https://github.com/NixOS/nixos-search/pull/1379#issuecomment-5005327009)
* Prior Discussions: https://github.com/NixOS/nixos-search/issues/1402, https://github.com/NixOS/nixos-search/issues/1183,
* Maintainers CC: @Malix-Labs, @KiaraGrouwstra, @raboof

Malix-Labs:
flake-info also have things that we cannot really replace, such as features for elasticsearch. however, it also have things that can be done in pure nix or at least with dedicated existing solutions in the nix ecosystem, and that imo is what we should be using

raboof:
How can I invoke the 'working prototype' to observe the claimed benefits?

Malix-Labs:
[...]

PR #1450, opened 2026-07-24
flake-info: modernize and simplify

Malix-Labs:
## Related

- [ ] #1434

### DIFF
--- a/flake-info/assets/commands/flake_info.nix
+++ b/flake-info/assets/commands/flake_info.nix
@@ -359,77 +359,38 @@ let
 
   collectSystems =
     outputKey: list:
-    lib.lists.foldr (
-      drv@{ attribute_name, system, ... }:
-      set:
-      let
-        # Check if this is a lightweight package entry (only has attribute_name, system, entry_type)
-        # Apps are not lightweight - they have bin/type fields
-        isLightweightPackage = !(drv ? name) && drv.entry_type == "package";
-
-        # For apps, check if they have metadata (bin/type fields)
-        isApp = drv.entry_type == "app";
-        appHasMetadata = isApp && (drv ? bin || drv ? type);
-
-        # Get existing entry or create new base
-        present =
-          set."${attribute_name}" or (
-            if isLightweightPackage then
-              {
-                platforms = [ ];
-                entry_type = "package";
-                inherit attribute_name;
-              }
+    let
+      grouped = lib.groupBy (x: x.attribute_name) list;
+      mergeEntry =
+        attribute_name: entries:
+        let
+          firstWithMeta =
+            let
+              found = lib.findFirst (e: e ? name || e ? bin) null entries;
+            in
+            if found != null then
+              found
             else
-              ({ platforms = [ ]; } // drv)
-          );
-
-        # Check if present entry has metadata
-        presentHasMetadata =
-          present ? name || (present.entry_type == "app" && (present ? bin || present ? type));
-
-        # Merge entries
-        drv' =
-          if isLightweightPackage then
-            if presentHasMetadata then
-              # Present has metadata, just add platform
-              present
-              // {
-                platforms = present.platforms ++ [ system ];
-              }
-            else
-              # Present lacks metadata, this must be a platform-specific package
-              # Evaluate it from this system
               let
-                metadata = evaluatePackageFromSystem outputKey system attribute_name;
+                firstEntry = lib.head entries;
+                meta = evaluatePackageFromSystem outputKey firstEntry.system attribute_name;
               in
-              if metadata != null then
-                present
-                // metadata
-                // {
-                  platforms = present.platforms ++ [ system ];
-                }
-              else
-                # Evaluation failed, just add platform without metadata
-                present
-                // {
-                  platforms = present.platforms ++ [ system ];
-                }
-          else
-            # Current entry has full metadata (package with name, or app with bin/type)
-            present
-            // drv
-            // {
-              platforms = present.platforms ++ [ system ];
-            };
+              if meta != null then firstEntry // meta else firstEntry;
 
-        drv'' = removeAttrs drv' [ "system" ];
-      in
-      set
-      // {
-        ${attribute_name} = drv'';
-      }
-    ) { } list;
+          rawPlatforms = lib.unique (map (e: e.system) entries);
+          targetOrder = [
+            "x86_64-linux"
+            "x86_64-darwin"
+            "aarch64-linux"
+            "aarch64-darwin"
+          ];
+          platforms =
+            (lib.filter (x: lib.elem x rawPlatforms) targetOrder)
+            ++ (lib.filter (x: !(lib.elem x targetOrder)) rawPlatforms);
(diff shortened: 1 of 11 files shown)


flake-info/src/commands/nixpkgs_info.rs
```diff
@@ -142,16 +141,16 @@ pub fn get_nixpkgs_package_services(nixpkgs: &Source) -> Result<HashMap<String,
 }
 
 pub fn get_nixpkgs_programs(nixpkgs: &Nixpkgs) -> Result<HashMap<String, HashSet<String>>> {
-    let mut command = Command::new("nix-instantiate");
```
KiaraGrouwstra, line 145: if v2 cli sufficed, why switch it to the flake cli, is that due to eval cache (and if so, does that even apply with `--impure`)?

Malix-Labs: `flake-info` already uses the nix3 cli throughout the rest of the codebase, so replacing legacy `nix-instantiate` keeps invocations uniform

This one is a direct 1:1 replacement with an added benefit since `nix eval --raw` prints the evaluated file path directly to stdout without the past required extra json serialization and`serde_json::from_str` unquoting overhead required in `nix-instantiate --json`

Also, even with `--impure`, i think `nix eval` take into account the internal evaluation cache `~/.cache/nix/eval-cache/` across runs, whereas `nix-instantiate` parses and evaluates the expression tree from scratch on every call

KiaraGrouwstra:
CI red still

Malix-Labs:
I did some more changes so it was back on red yes, i'll resolve

flake.nix
```diff
@@ -9,193 +9,219 @@
   };
 
   inputs = {
-    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
-    flake-utils.url = "github:numtide/flake-utils";
+    nixpkgs.url = "https://channels.nixos.org/nixos-unstable/nixexprs.tar.xz";
+    flake-parts = {
+      url = "github:hercules-ci/flake-parts";
```
KiaraGrouwstra, line 14: what prompted this particular switch?

Malix-Labs: flake-parts because it is the successor of flake-utils

nixpkgs because of https://discourse.nixos.org/t/questions-about-inputs-nixpkgs/69095/14?u=malix (the whole thread is interesting)

KiaraGrouwstra: `flake-utils` would seem like the more light-weight option, lacking a dependency on nixpkgs.
if we did (and do) not need functionality from `flake-parts` then, what would be concrete benefits of switching?

flake-info/assets/commands/flake_info.nix
```diff
@@ -247,70 +249,58 @@ let
...
-        in
-        (builtins.head entries)
-        // {
-          service_package = builtins.head packages;
-          service_packages = packages;
-        };
-    in
-    lib.mapAttrsToList (_: mergeGroup) grouped;
+    opts
+    |> lib.groupBy (
```
KiaraGrouwstra: afaik pipe operators differed between nix implementation (at least original nix vs lix), so i would avoid using them in published code that might (somehow) have/get downstream consumers

Malix-Labs: okay, I'll remove it

flake.nix
Malix-Labs: Flake-parts has a dependency for nixpkgs-lib, which is only the lib part of nixpkgs, it's lightweight in that sense;
Making it follow the full nixpkgs (since we need this one) makes it not have any other transitive deps

It is handy for per-system, theoretically speaking
We can also make a dev partition and use git-hooks.nix or treefmt-nix since they have a built-in flakeModules

There are some other arguments in [Why you don't need flake-utils](https://ayats.org/blog/no-flake-utils)

flake-info/assets/commands/flake_info.nix
Malix-Labs: Done in fad51666069c1c17609b64bbee117507007756aa

flake.nix
KiaraGrouwstra: > We can also make a dev partition and use git-hooks.nix or treefmt-nix since they have a built-in flakeModules

is this about `flake-parts` vs `flake-utils`? i don't quite follow.

i think adding `flake-parts` would further tie us up into handling dependency management by flake inputs, which i'm not quite sure is (or should be) where the ecosystem is going.

Malix-Labs: > is this about `flake-parts` vs `flake-utils`? i don't quite follow

It's about the benefits flake-parts would have over flake-utils

flake-utils was already what we used in that flake, I simply replaced it with flake-parts

> i think adding `flake-parts` would further tie us up into handling dependency management by flake inputs, which i'm not quite sure is (or should be) where the ecosystem is going.

Do you have any alternative in mind?

KiaraGrouwstra: thanks, i understand what you're saying now w.r.t. doing a dev flake using `flake-parts`.

i tend to reconcile these by making `flake.nix` a [thin wrapper](https://git.fediversity.eu/fediversity/nix-module-form/src/branch/main/flake.nix) (preventing need for `flake-utils`) of a vanilla entry-point handling dependencies otherwise, such as using npins. that way, dependencies are already handled lazily, preventing need for workarounds to issues specific to flake inputs such as lazy trees and `flake-parts`' dev flakes.

to be fair, integrating such dependency management with flake input overrides involves [creative use of dummy inputs](https://codeberg.org/kiara/poc-override-nix-deps/src/branch/main/flake.nix), tho i'm under the impression flake input overrides in practice tend to function less as a reason people use flakes, but rather as a way flake users cope with flake inputs.

to be fair, i suppose if anything, external consumers seem at best a secondary goal of this repo, so perhaps for our purposes `flake-parts` might do as well - i imagine our `flake.nix` in this repo in practice is more aimed at integrating with use of tools oriented at nix development, rather than toward external consumers.

KiaraGrouwstra (approved):

raboof:
Did this break `.github/workflows/check-flake-files.yml`? https://github.com/NixOS/nixos-search/actions/runs/30210290176/job/89815246351?pr=1465

@Malix-Labs could you have a look at that?

raboof:
@KiaraGrouwstra had a possible fix in #1467 , let's see if that fixes #1465

Malix-Labs:
@raboof I don't remember having such issues on my side after da7cbc1dc6567578604121295480acda94587b2e and the other related commits

I'll monitor if #1467 fixes this, but I'll log off soon

---

It seems that https://github.com/NixOS/nixos-search/actions/runs/30210290176/job/89815246351?pr=1465 ran for less than 3 minutes

And after #1467 as merged and you force pushed, https://github.com/NixOS/nixos-search/actions/runs/30213105249/job/89822562638?pr=1465 is running for 18 minutes now, so @KiaraGrouwstra might have fixed it

Also see my comment in https://github.com/NixOS/nixos-search/pull/1467#issuecomment-5084716950

PR #1457, opened 2026-07-25
frontent/search: use repology popularity for search ranking

hsjobeki:
This should improve the ranking by boosting up popular packages.


KiaraGrouwstra (approved):
slight score boost huh, i wonder if this could be tweaked still, but let's say it's better than the status quo

PR #1458, opened 2026-07-25
frontend/search: rank by reverse-dependency count

hsjobeki:
Depends on #1457 

The current benchmark cannot really measure this impact.
It is most visible on libraries. We should add more library queries to the benchmark.


KiaraGrouwstra:
thanks, the other one is merged so this one can be rebased now.

content-wise, looks like intent-based queries get a boost here, so that seems good.

hsjobeki:
@KiaraGrouwstra just rebased

KiaraGrouwstra (approved):

KiaraGrouwstra:
would've been nice to see #1461 land first, but to be fair logically speaking i wouldn't necessarily expect this to aggravate things much

PR #1459, opened 2026-07-25
frontend/search: split relevance benchmark into packages vs options

### LINKED ISSUE #1443: split query ranking benchmark into packages vs options
this suggestion by @raboof was taken from https://github.com/NixOS/nixos-search/pull/1441#issuecomment-5045643925, tho i'd previously discussed this with @hsjobeki as well.

while having one over-arching search would seem cool, our status quo is based on queries targeting a single category, which for benchmarking purposes would make things easier as well.

a cross-category ranking benchmark may be of interest as well, but could potentially be (re)added later as well.

KiaraGrouwstra:
The benchmark previously merged package-index and option-index hits into a single score-sorted list scored against a `relevant` array mixing `pkg:` and `opt:` ids. That merge masked package-search problems: a query with both kinds of relevants could look healthy on the merged score while package ranking was quietly broken (e.g. `continuwuity` returns good option hits but poor package hits). The search pages target a single category anyway, so the benchmark now mirrors that and scores package search and option search separately, each with its own metrics.

`queries.json` is split into `queries-packages.json` (98 queries, `pkg:` relevants) and `queries-options.json` (119 queries, `opt:` relevants); queries with both kinds appear in both files carrying only that file's relevants. `run.mjs` runs each file against only its index via `scoreTrack`, replaces `mergeHits` with a single-index `rankHits`, and emits a two-section report with a combined per-query table carrying a `track` column.

Also tidies g136's option relevant from the templated `opt:services.redis.servers.<name>.enable` to the concrete `opt:services.redis.servers` for consistency with g084. This does not change g136's score -- the semantic query itself is the limiter, not the relevant.

A cross-category ranking benchmark can be re-added later.

Closes #1443.

Assisted-by: Claude:claude-fable-5

### DIFF
--- a/frontend/benchmark/queries-options.json
+++ b/frontend/benchmark/queries-options.json
@@ -1,106 +1,51 @@
 [
     {
-        "id": "g001",
-        "q": "ripgrep",
-        "category": "exact",
-        "relevant": ["pkg:ripgrep"]
-    },
-    {
-        "id": "g002",
-        "q": "htop",
-        "category": "exact",
-        "relevant": ["pkg:htop"]
-    },
-    {
-        "id": "g003",
-        "q": "jq",
-        "category": "exact",
-        "relevant": ["pkg:jq"]
-    },
-    {
-        "id": "g004",
-        "q": "tmux",
-        "category": "exact",
-        "relevant": ["pkg:tmux"]
-    },
-    {
-        "id": "g005",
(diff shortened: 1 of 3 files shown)


hsjobeki:
Ah i was about to do the same change :+1:

raboof (approved):

PR #1464, opened 2026-07-26
feat(flake-info): index the modular-service base/environment split

KiaraGrouwstra:
At nixpkgs, I have proposed (NixOS/nixpkgs#540863) splitting each modular service into two halves: a portable package service at `pkgs/**/service.nix`, reachable as `pkgs.<pkg>.services.<svc>`, and environment-specific implementations thereof (so far: a NixOS system-level service) at `nixos/modules/system/service/modular/<pkg>/<svc>/system.nix`.
Both are enumerated in a registry keyed `<environment>.<pkg>.<service>` and exposed as `lib.nixos.modularServices` (all environments) and `config.modularServices` (the `system` slice).

The current change then helps validate that the proposed method may help make discoverable what services would be supported on what environments (NixOS/nixpkgs#545204).

image

image

(Note that, while in the screenshots the maintainers listed for the package service vs its NixOS system implementation are identical, these are separate - one does not need to maintain one just because they volunteered to maintain the other.)

### DIFF
--- a/flake-info/src/commands/nixpkgs_info.rs
+++ b/flake-info/src/commands/nixpkgs_info.rs
@@ -82,8 +82,9 @@ pub fn get_nixpkgs_info(
                 .into_iter()
                 .collect();
             let modular_services = package_services.remove(&attribute).unwrap_or_default();
-            let modular_service_imports =
-                package_service_imports.remove(&attribute).unwrap_or_default();
+            let modular_service_imports = package_service_imports
+                .remove(&attribute)
+                .unwrap_or_default();
             let dep_count = dep_counts.get(&attribute).copied();
             let repology_repos = repology_counts.get(&attribute).copied();
             NixpkgsEntry::Derivation {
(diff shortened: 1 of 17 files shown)


flake-info/src/commands/nixpkgs_info.rs
```diff
@@ -124,22 +130,39 @@ fn resolve_repology_counts(file: &Option<PathBuf>) -> HashMap<String, u64> {
...
     let de = &mut serde_json::Deserializer::from_str(output);
-    let map: HashMap<String, Vec<String>> = serde_path_to_error::deserialize(de)
-        .with_context(|| "Could not parse package-services map")?;
+    let map: HashMap<String, T> = serde_path_to_error::deserialize(de)
+        .with_context(|| format!("Could not parse {}", attribute))?;
     Ok(map)
 }
 
+pub fn get_nixpkgs_package_services(nixpkgs: &Source) -> Result<HashMap<String, Vec<String>>> {
+    get_package_map(nixpkgs, "nixos-package-services")
```
KiaraGrouwstra: i dunno if this is supposed to match anything, but should this semantically not be closer to nixpkgs package services?

PR #1472, opened 2026-07-26
fix(frontend): improve terminal theming

Sigmanificient:
I couldn't make the terminal light theme, so i default to use the branding colors. Cant really tell if this is better.

### before

image

### after

image

### DIFF
--- a/frontend/src/Page/Options.elm
+++ b/frontend/src/Page/Options.elm
@@ -63,7 +63,7 @@ import Search
         , decodeResolvedFlake
         )
 import Search.Query
-import SyntaxHighlight exposing (elm, oneDark, toBlockHtml, useTheme)
+import SyntaxHighlight exposing (nix, toBlockHtml)
 import Task
 import Url
 import Utils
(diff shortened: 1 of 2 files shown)


KiaraGrouwstra:
i'm curious as to how others feel here - i'm not sure i find it nicer to look at, fwiw

Sigmanificient:
Maybe with a more blueish bg? 

image

KiaraGrouwstra:
i feel like i might just have like pastel-on-brown sensibilities, and perhaps i'm just not the best person to ask here.
for what it's worth, blue light seems less great to look at when it's dark, so maybe i've just been avoiding that more in general, tho i'm having trouble saying which would otherwise look prettier.
that said, i'm also not very sure how that type of reasoning might go with our marketing style guide - tho that could also still just be not fully defined yet, if so <https://nixos.org/> otherwise mostly just did light theme, still.

raboof:
> Cant really tell if this is better.

I don't have a preference. Is this something that could be fleshed out 'upstream' (also for the site, the docs, etc) and then we can follow that lead here?

KiaraGrouwstra:
i'm not sure use of nixos branding colors would be as relevant for upstream, per se

raboof:
sorry for being unclear, by 'upstream' here I meant @Sigmanificient and others working on NixOS design/styling more generally, not just for the search

PR #1486, opened 2026-07-27
evalFlake: init

Malix-Labs:
Rewrite flake_info.nix (with an arbitraty legacy schema) in evalFlake.nix around flake-schemas

Issue #1182, opened 2026-04-08
feat(flakes): rev tracking


Malix-Labs:
## Currently

[flakes](https://github.com/NixOS/nixos-search/blob/main/flakes)'s latest valid revision isn't tracked :

- If one flake breaks, the whole batch doesn't get updated in elasticsearch
  - Current broken flakes are commented out as a bandaid workaround
- We are not able to add a state information like "✔️ Evaluable" / "❌ Unevaluable" or similar to each flake along with the rev and date of the latest check

## Solution

We should feature rev tracking in elasticsearch's ingestion layer to enable and fix those elements

Issue #1116, opened 2025-12-08
Make it more obvious when flake evaluations fail


raboof:
The hourly import currently succeeds even when there's individual flakes that failed. The PR validation does fail when any fails.

We should probably make the import job run to completion when one flake fails, but then fail with a clear report/issue, so we notice when flakes stop working, rather than only noticing when an unrelated PR fails

PR #1489, opened 2026-07-28
flake-info: fix the evaluation bugs behind the failing manual group import

### LINKED ISSUE #1487: Failing import for manual group
This issue was triggered by https://github.com/NixOS/nixos-search/actions/runs/30260562793

KiaraGrouwstra:
Closes #1487.

Fixes three bugs in `flake-info` -- not in the flakes -- behind the failing `manual` group imports. They were root-caused from run 30222963951 (#1482), where five of ~120 flakes failed; #1487 is the still-open recurrence, where `codeberg.org/alch_emi/embers-nix-packages` fails the whole group on `missing field 'name'`. That is the first bug below, and the flake imports cleanly here at its current `main`.

A failing attribute now amputates only itself: an attribute that throws is dropped, a system whose output set throws is skipped, and everything else in the flake is still indexed. Recording partial-versus-complete evaluation state in ElasticSearch is #1182 and is not attempted here.

## Name-less package entries reach `serde`

`readSchemaItems` evaluates full metadata only on `referenceSystem`, where `evalDrvMetadata` returns `null` for non-derivations and the entry is dropped. For every other system it emits a name-less stub with no derivation check, and `collectSystems` keeps that stub when the reference system yielded no metadata either. `FlakeEntry::Package` requires `name`, so a single non-derivation under `packages` fails deserialization for the whole flake.

`mergeEntry` now returns `null` for a merged package with no `name`, and `collectSystemEntries` filters those out. `collectSystems` is the only producer of final package entries, so no name-less package can reach `serde` regardless of which branch produced the stub. Apps have no `name` by design, so the check is restricted to `entry_type == "package"`.

Hit by `git+https://codeberg.org/alch_emi/embers-nix-packages.git?ref=main`, whose `generate-caddy-sidecars` is a plain attrset.

## Forcing a per-system inventory node is unguarded

`readSchemaItems` forces `sysNode.children`, which evaluates the flake's entire output set for that system. A system nixpkgs no longer supports therefore aborts evaluation of every other system too. #1485 guards the individual entries but not this. The forcing now goes through `safeEval`, skipping just that system.

Hit by `github:nix-community/fenix`, and by `github:StardustXR/telescope` on `riscv64-linux`.

## `nix` stderr never reaches the error message

`nix_eval_command` sets `log_output_on_error`, so stderr goes only to the log; `command_run`'s error renders as just `command '...' failed: exit status: 1`. That is what `FlakeInfoError::Group` aggregates and what the import workflow pastes into the auto-filed issue, which is why #1318 and #1260 were closed as stale rather than fixed -- four of the five entries in #1482 carry no diagnostic information at all, and #1487 needed the raw 18 MB job log to identify the flake at fault.

`run_capturing_stderr` disables `command_run`'s check, and on a non-zero exit builds the error from the exit status plus the last 40 lines of stderr (traces run to thousands of lines, and the message is pasted verbatim into an issue). All `nix` call sites are routed through it. Logging is unchanged. Progress on #1116.

## Re-enabled flakes

`flakes/manual.toml` had `hannesgith/neohtop` commented out for `missing field 'name'` and `nix-community/fenix` for `x86_64-darwin throws deprecated` -- exactly the two bugs above. Both import cleanly now. The other commented-out entries were disabled for unrelated reasons and are left alone.

## Testing

`nix build .#checks.x86_64-linux.flake-info` passes. The `hydra`, `agenix` and `deploy-rs` fixtures pin flake revisions and are byte-identical, confirming no behaviour changed for flakes that already worked. `basic-flake` gains a non-derivation package (dropped), a throwing system (skipped), an app that throws everywhere (dropped) and an app that throws only on darwin (kept, listing only the platforms it evaluated on).

The five failing flakes from run 30222963951 all evaluate, as do the two re-enabled ones, and no package entry lacks `name`:

| flake | entries | packages | apps |
|---|---|---|---|
| `embers-nix-packages` | 10 | 10 | 0 |
| `deadnix` | 3 | 2 | 1 |
| `lnbits-legend` | 5 | 2 | 3 |
| `telescope` | 7 | 4 | 3 |
| `mcp-nixos` | 5 | 3 | 2 |
| `neohtop` | 3 | 3 | 0 |
| `fenix` | 139 | 139 | 0 |

A full `--json group ./flakes/manual.toml manual --report` run over all ~120 flakes exits 0 and writes no `report.txt`, yielding 38372 documents -- 36405 options, 1920 packages and 47 apps.

## Relationship to #1486

#1486 restructures `flake_info.nix` into a flake with pinned inputs and a `lib`, and rebases the Rust side onto it. The two do not duplicate each other: it keeps the per-entry `tryEval` from #1485 but leaves `sysNode.children` unguarded (and for `legacyPackages` additionally forces `resolved.legacyPackages.<system>` through `walkAttrTree`, widening the exposure), has no name-less-package check, and does not touch the stderr path. Both fixes here apply on top of that restructure. Whichever lands first, the other rebases -- happy to go second, since this one is small and #1486 is still WIP.

## Out of scope

Three things surfaced while investigating that belong in separate issues:

- `mergeEntry` picks the alphabetically-first system's `bin` rather than `referenceSystem`'s, so `deploy-rs` indexes its `aarch64-darwin` program while the fixture records the `x86_64-linux` path. It only still matches because `rmDerivationHash` strips the store hash. Reordering by the `targetOrder` list already in `flake_info.nix` would make it deterministic with no fixture change.
- Apps have never been rendered by the frontend -- the flake page filters to `package` and `option`, and `app_bin` appears only in backend files. Whether to index them at all is worth asking.
- Once a flake can be partially evaluated, the error label wants to say which flake failed and that it was `flake_info.nix` that failed on it, and ElasticSearch wants the pass/fail state alongside the data (#1182).

### DIFF
--- a/flake-info/assets/commands/test/basic-flake/flake.nix
+++ b/flake-info/assets/commands/test/basic-flake/flake.nix
@@ -15,47 +15,59 @@
       forAllSystems = nixpkgs.lib.genAttrs systems;
     in
     {
-      packages = forAllSystems (
-        system:
-        let
-          pkgs = import nixpkgs { inherit system; };
-        in
-        {
-          # Package available on all systems
-          hello = pkgs.writeShellScriptBin "hello" ''
-            echo "Hello from all systems"
-          '';
-
-          # Package available on all systems with metadata
-          test-package = pkgs.stdenv.mkDerivation {
-            pname = "test-package";
-            version = "1.2.3";
-            src = pkgs.writeText "test.txt" "test content";
-            dontUnpack = true;
-            installPhase = ''
-              mkdir -p $out
-              cp $src $out/test.txt
+      packages =
+        forAllSystems (
+          system:
+          let
+            pkgs = import nixpkgs { inherit system; };
+          in
+          {
+            # Package available on all systems
+            hello = pkgs.writeShellScriptBin "hello" ''
+              echo "Hello from all systems"
             '';
-            meta = with pkgs.lib; {
-              description = "A test package";
-              longDescription = "This is a longer description for testing";
-              license = licenses.mit;
+
+            # Package available on all systems with metadata
+            test-package = pkgs.stdenv.mkDerivation {
+              pname = "test-package";
+              version = "1.2.3";
+              src = pkgs.writeText "test.txt" "test content";
+              dontUnpack = true;
+              installPhase = ''
+                mkdir -p $out
+                cp $src $out/test.txt
+              '';
+              meta = with pkgs.lib; {
+                description = "A test package";
+                longDescription = "This is a longer description for testing";
+                license = licenses.mit;
+              };
             };
-          };
-        }
-        // pkgs.lib.optionalAttrs pkgs.stdenv.isDarwin {
-          # Darwin-only package
-          darwin-specific = pkgs.writeShellScriptBin "darwin-test" ''
-            echo "Darwin only"
-          '';
-        }
-        // pkgs.lib.optionalAttrs pkgs.stdenv.isLinux {
-          # Linux-only package
-          linux-specific = pkgs.writeShellScriptBin "linux-test" ''
-            echo "Linux only"
-          '';
-        }
-      );
+
+            # Not a derivation, so it has no `name` -- must be dropped rather
+            # than emitted as a name-less package entry.
+            not-a-derivation = {
+              some = "attrset";
+            };
+          }
+          // pkgs.lib.optionalAttrs pkgs.stdenv.isDarwin {
+            # Darwin-only package
+            darwin-specific = pkgs.writeShellScriptBin "darwin-test" ''
+              echo "Darwin only"
+            '';
+          }
+          // pkgs.lib.optionalAttrs pkgs.stdenv.isLinux {
+            # Linux-only package
+            linux-specific = pkgs.writeShellScriptBin "linux-test" ''
+              echo "Linux only"
+            '';
+          }
+        )
+        // {
+          # A system whose entire package set throws -- that system must be
+          # skipped rather than aborting evaluation of the whole flake.
+          riscv64-linux = throw "this system is not supported";
+        };
 
       apps = forAllSystems (
         system:
@@ -66,6 +78,24 @@
           app-hello = {
             type = "app";
             program = pkgs.hello.outPath + "/bin/hello";
+          };
+
+          # Throws on every system: the attribute is dropped rather than
+          # aborting evaluation of the whole flake.
+          throwing-app = {
+            type = "app";
+            program = throw "this app cannot be evaluated";
+          };
+
+          # Throws only on darwin: the attribute survives, listing just the
+          # platforms it evaluated on.
+          partly-throwing-app = {
+            type = "app";
+            program =
+              if pkgs.stdenv.isDarwin then
+                throw "this app is unsupported on darwin"
+              else
+                pkgs.hello.outPath + "/bin/hello";
           };
         }
       );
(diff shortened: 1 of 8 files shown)


Malix-Labs:
@KiaraGrouwstra 

I think this would close #1487 instead of #1482 (I haven't double checked, but I think the error was different between the two, otherwise it was a previous one)

also see #1486, which tries to implement a fix for #1490 (notably. I also updated the test cases, and made some inputs pinned in a flake.lock, then made flake-info consume call the flake.nix's lib.

Note that it is currently a WIP

Also, I think we should implement it so that a failing attribute only amputate that attribute to the ElasticSearch update (keeping the specific attribute previously passing evaluation as is)
For that, the code would need to be way more modular, allowing each leaf to be amputated as scoped and as cleanly as possible (and if a single leaf has been amputated, consider the flake eval to be partial only)

Some of those issues, in my opinion, stems from adhering to a sub-optimal expected output format : https://github.com/NixOS/nixos-search/blob/main/flake-info/assets/commands/test/expected-outputs/basic-flake.json
Does not seem right to me

Then, we should implement #1182

KiaraGrouwstra:
@Malix-Labs i cannot tag you as reviewer yet, but your review would be welcome :), i think our work mostly seems complementary.

flake-info/assets/commands/test/basic-flake/flake.nix
```diff
@@ -15,47 +15,59 @@
...
-          # Linux-only package
-          linux-specific = pkgs.writeShellScriptBin "linux-test" ''
-            echo "Linux only"
-          '';
-        }
-      );
+
+            # Not a derivation, so it has no `name` -- must be dropped rather
+            # than emitted as a name-less package entry.
+            not-a-derivation = {
```
Malix-Labs, line 49: good idea

flake-info/assets/commands/test/basic-flake/flake.nix
```diff
@@ -15,47 +15,59 @@
...
+            # Linux-only package
+            linux-specific = pkgs.writeShellScriptBin "linux-test" ''
+              echo "Linux only"
+            '';
+          }
+        )
+        // {
+          # A system whose entire package set throws -- that system must be
+          # skipped rather than aborting evaluation of the whole flake.
+          riscv64-linux = throw "this system is not supported";
```
Malix-Labs, line 69: this one too, good idea

builtins.tryEval in this Nix version doesn't catch MissingAttrError (only  throw / assert ), so `(tryEval (isDerivation val)).value`  escapes when val is a thunk whose construction throws, this was the case in my PR for agenix with `pkgs.google-chrome-beta` for example

The fix was to force val to WHNF with a separate tryEval before doing anything with it

flakes/manual.toml
Malix-Labs, line 1: have you tried running your PR against those flakes?

Malix-Labs:
> Whether to index them [apps] at all is worth asking

Wouldn't it be worth to display them in the frontend?

> happy to go second

I prefer that this PR gets merged first, I have a bigger commit to do which uses pinned inputs when preferable and exposes the functions under `lib` instead

flakes/manual.toml
KiaraGrouwstra: yes! these did in fact suffer from the same classes of failure as the new ones

KiaraGrouwstra:
fair enough! and yes if we could expose `apps` then great

Malix-Labs:
I don't really know if it would be worth it, but it's not that hard to add

flakes/manual.toml
Malix-Labs: @KiaraGrouwstra I'm not sure I understand what you mean;

by "these", do you mean "all", and by "new ones", you mean the newly uncommented ones?

do you mean that this PR resolves the bug that only the previously commented flakes had?

Malix-Labs:
Bruh, exporting app was actually already done but is broken because we didn't make it pass through a serde rename, I have fixed it on my local branch for now

Malix-Labs:
@raboof can you merge this one?

Malix-Labs (approved):

raboof (approved):
thanks @Malix-Labs for the review

PR #1500, opened 2026-08-06
Benchmark: add precision (RBP)

hsjobeki:


frontend/benchmark/run.mjs
```diff
@@ -180,10 +180,11 @@ function recallAtK(ranked, relevant, k) {
 //   `1 / (1 - p)` results.
 //
 // Requires `"exhaustive": true`, meaning `relevant` enumerates every acceptable
-// answer; otherwise a good-but-unlisted hit scores as noise. Returns `null` when
-// there are no hits, which drops the query from the mean.
+// answer; otherwise a good-but-unlisted hit scores as noise. A page with no hits
+// scores 0 -- it found none of the answers -- so that a query which breaks
+// outright cannot leave the mean by dropping out of it.
```
KiaraGrouwstra: is 'punish 0-hit results' really what we want here?
as in, if the metric's point was to punish noise (as per #1442), then why should we assign the worst possible score to a result of _zero_ noise?
because, is your consideration of "it found none of the answers" not already covered by like, every single other one of our metrics, already?
in that sense, (while the `matchedExact` thing here is probably fine), is this particular change not making it worse than having this metric opt ouf of the mean, in that particular scenario?

hsjobeki: Yeah I agree. We still have the other metrics, need to review them together.

I was intially worried that the metrics could be gamed.

KiaraGrouwstra (approved):

PR #1512, opened 2026-08-12
frontend/search: rank the enable option of a module first

hsjobeki:
A search for a module returned the wrong option.
`services.nginx` put `services.nginx.additionalModules` first. It left `services.nginx.enable` outside the first ten.

Most modules have one `enable` option.
The user wants that option.
1635 of the 1862 two-segment module paths have one. That is 92% under `services` and 97% under `programs`.

`option_name.attr_path_reverse` holds every suffix of a dotted path. So one exact term matches `services.nginx` and a bare `nginx`. A weaker second clause lifts every `enable` option. A partial name such as `graf` needs that clause.

The benchmark ran on `latest-50-nixos-unstable` at k=10. Options Success@10 rises 0.538 -> 0.681.
MRR rises 0.286 -> 0.519.
Recall@10 rises 0.534 -> 0.676.
The dotted and exact categories reach Success 1.000. Prefix rises 0.632 -> 0.947.
RBP holds at 0.806.
The package request body does not change.

### DIFF
--- a/frontend/src/Search/Query.elm
+++ b/frontend/src/Search/Query.elm
@@ -96,6 +96,7 @@ packagesBody query from size sort selectedBuckets =
         ]
         [ "package_attr_name", "package_pname", "package_programs" ]
         [ "package_description^3", "package_longDescription^1" ]
+        Nothing
         [ { field = "package_repology_repos", pivot = 20.0 }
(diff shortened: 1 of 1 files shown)


KiaraGrouwstra (approved):

KiaraGrouwstra:
options ranking benchmark go brr

PR #1513, opened 2026-08-12
Benchmark: score option queries against the whole module cluster

hsjobeki:
This PR Adds more option queries. 
I went on a journey through nixos dicsourse and some more forums stated below to rate what people usually want to find.

---- 

Option relevance comes from counting mentions in real user text. Four corpora were read in full:
949 wiki.nixos.org pages through the MediaWiki API, 597 Discourse threads through /raw/<topic>,
2625 Reddit and StackExchange pages through their APIs, and 554 issues plus 925 pull requests from this repo.

Each corpus ranks services.nginx.virtualHosts above services.nginx.enable. The wiki counts are 43 against 31.
The Discourse counts are 21 against 14.
So 28 module queries now expect an evidence-ranked option cluster. Those queries carry "exhaustive", which prices page junk into RBP.

37 new "scoped" queries pair a service with option words. Examples are "nginx config", "git config" and "postgresql data dir". Leaf names repeat across services, so one shape covers many modules. The wiki shows "enable" on 268 services and "settings" on 83. This shape scores 0.432 Success@10 and 0.080 RBP today. "postgresql data dir" and "syncthing data dir" return identical pages.

A terms query checked all 249 expected names against latest-50-nixos-unstable. The check rejected ten candidate names.
hardware.opengl.enable became hardware.graphics.enable in 24.11. Redis configuration lives under services.redis.servers.<name>.

The by-category table now prints Recall@K and RBP. Cluster expectations move those two metrics.

### DIFF
--- a/frontend/benchmark/queries-options.json
+++ b/frontend/benchmark/queries-options.json
@@ -3,49 +3,136 @@
         "id": "g008",
         "q": "nginx",
         "category": "exact",
-        "relevant": ["opt:services.nginx.enable"]
+        "exhaustive": true,
+        "relevant": [
+            "opt:services.nginx.enable",
+            "opt:services.nginx.virtualHosts",
+            "opt:services.nginx.user",
+            "opt:services.nginx.group",
+            "opt:services.nginx.package",
+            "opt:services.nginx.config",
+            "opt:services.nginx.httpConfig",
+            "opt:services.nginx.appendHttpConfig",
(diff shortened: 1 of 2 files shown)


KiaraGrouwstra (approved):
the data collection method sounds a bit roundabout since we may also just be able to query ElasticSearch - but thanks for fleshing out our benchmark!

PR #1514, opened 2026-08-12
Update queries-packages.json

jerrygreen:
Add nodejs

> mentioned in https://github.com/NixOS/nixos-search/issues/1438#issuecomment-5272436536

### DIFF
--- a/frontend/benchmark/queries-packages.json
+++ b/frontend/benchmark/queries-packages.json
@@ -739,5 +739,19 @@
         "category": "exact",
         "exhaustive": true,
         "relevant": ["pkg:matrix-continuwuity"]
+    },
+    {
+        "id": "g152",
+        "q": "node",
+        "category": "exact",
+        "exhaustive": true,
+        "relevant": [
+            "pkg:nodejs",
+            "pkg:nodejs_latest",
+            "pkg:nodejs_26",
+            "pkg:nodejs_24",
+            "pkg:nodejs_22",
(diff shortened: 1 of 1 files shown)


frontend/benchmark/queries-packages.json
```diff
@@ -739,5 +739,19 @@
         "category": "exact",
         "exhaustive": true,
         "relevant": ["pkg:matrix-continuwuity"]
+    },
+    {
+        "id": "g152",
+        "q": "node",
+        "category": "exact",
+        "exhaustive": true,
```
KiaraGrouwstra, line 747: this would make the claim not just that `nodejs-slim` has _less_ relevance to query `node`, but in fact that it is _not significantly relevant_ to the query, nor any other (unlisted) result.

i realize that would actually seem a cope with our current benchmarking mechanism to express your feedback at https://github.com/NixOS/nixos-search/issues/1438#issuecomment-5271848270, tho i think it we could probably do better than you suggested here still. i filed that at #1515, while i'll just accept this change for now.

KiaraGrouwstra (approved):

frontend/benchmark/queries-packages.json
hsjobeki: In fact it seems you need fuzzy matching for this. `node` != `nodejs`. Without fuzzy matching you'd get `0` results or only those matching `node` exactly.

PR #1518, opened 2026-08-13
Benchmark: score the gold set as ranked tiers

### LINKED ISSUE #1515: ranking benchmarking ordinal relevance
while we have recently started benchmarking our query ranking algorithm based on whether specified results were ranked (highly) for a given query, https://github.com/NixOS/nixos-search/issues/1438#issuecomment-5271848270 introduced a new concern: as users, we have existing notions not only as to _whether_ given results are relevant to a query, but _also_ of which results are _more_ relevant to a given query.

in order to be able to judge our ranking algorithm to incorporate such user expectations, we may need to then include information like this in our ranking benchmark as well, somehow.

KiaraGrouwstra:
Add two benchmarking metrics that read the gold set as *ranked* rather than as a flat set:

- **`BestRR`**, the reciprocal rank of the one answer the query most wants. It collapses to `MRR` on a single-answer query, so it only speaks up on the cluster queries it was added for. It reads **0.167** on the `node` page above.
- **`nDCG@10`**, which grades a hit in tier `i` of `T` at `T - i`, pricing the whole page against its ideal ordering rather than one position.

To express the ranking, `relevant` becomes a list of tiers, each tier a list of ids tied at that rank. Order *between* tiers is a claim, order *within* one is not: `[[a, b, c]]` states no preference, `[[a], [b], [c]]` is a strict order, and `[[a], [b, c]]` is a best answer followed by a tie. The best answer is the top tier's single id, or nothing at all where that tier is a tie - so ordering claims are only ever made where a curator wrote one down, which keeps them off the queries that cannot support one. `text editor` has no business ranking `emacs` over `vim`, and `container runtime` has no business ranking `docker` over `podman`. A query stating no preference drops out of the `BestRR` mean and keeps a flat gold set under `nDCG`, which is then ordinary binary nDCG.

The annotated groups are settled rather than judgment calls: packages where the query names a package and the gold set is that package plus its variants (`postgres` wants `postgresql`, `dcoker` wants `docker`), options where the query names a module and the gold set is its cluster with the `enable` option first - the thesis of #1512, and version families where the tiers are the releases, newest first (`nextcl` wants `nextcloud34`). Coverage lands at 85/99 package queries and 164/175 option queries. `intent` queries are left unannotated throughout.

| track | Success@10 | MRR | Recall@10 | RBP (p=0.8) | BestRR | nDCG@10 |
| packages (n=99) | 0.768 | 0.659 | 0.715 | 0.675 (n=22) | 0.698 (n=85) | 0.660 |
| options (n=175) | 0.663 | 0.513 | 0.483 | 0.301 (n=92) | 0.531 (n=164) | 0.433 |

Closes #1515.

### DIFF
--- a/frontend/benchmark/queries-options.json
+++ b/frontend/benchmark/queries-options.json
@@ -4,20 +4,21 @@
         "q": "nginx",
         "category": "exact",
         "exhaustive": true,
-        "best": "opt:services.nginx.enable",
         "relevant": [
-            "opt:services.nginx.enable",
-            "opt:services.nginx.virtualHosts",
-            "opt:services.nginx.user",
-            "opt:services.nginx.group",
-            "opt:services.nginx.package",
-            "opt:services.nginx.config",
-            "opt:services.nginx.httpConfig",
-            "opt:services.nginx.appendHttpConfig",
-            "opt:services.nginx.recommendedProxySettings",
-            "opt:services.nginx.recommendedTlsSettings",
-            "opt:services.nginx.recommendedGzipSettings",
-            "opt:services.nginx.recommendedOptimisation"
+            ["opt:services.nginx.enable"],
+            [
+                "opt:services.nginx.virtualHosts",
+                "opt:services.nginx.user",
+                "opt:services.nginx.group",
+                "opt:services.nginx.package",
+                "opt:services.nginx.config",
+                "opt:services.nginx.httpConfig",
+                "opt:services.nginx.appendHttpConfig",
+                "opt:services.nginx.recommendedProxySettings",
+                "opt:services.nginx.recommendedTlsSettings",
+                "opt:services.nginx.recommendedGzipSettings",
+                "opt:services.nginx.recommendedOptimisation"
+            ]
         ]
     },
     {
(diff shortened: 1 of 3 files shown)


frontend/benchmark/queries-options.json
```diff
@@ -4,6 +4,7 @@
         "q": "nginx",
         "category": "exact",
         "exhaustive": true,
+        "best": "opt:services.nginx.enable",
```
hsjobeki: Haven't fully wrapped my head around this PR yet. Seems best is strictly the head of relavant? 
How do you like the idea, where we take the order of relevant into account rather than rating one best?

KiaraGrouwstra: there are cases where we don't feel comfortable making claims on which one is best, so i instead now generalized it to at tier-based system

frontend/benchmark/run.mjs
```diff
@@ -341,14 +462,22 @@ const perQuery = [
 const lines = [
     "# Relevance benchmark: frontend query vs deployed ES",
     "",
-    `> Index: \`${INDEX}\`. metrics: (Success@${K}, MRR, Recall@${K}, RBP(p=${P})).`,
+    `> Index: \`${INDEX}\`. metrics: (Success@${K}, MRR, Recall@${K}, RBP(p=${P}), BestRR, nDCG@${K}).`,
     "",
     `> \`RBP\` is rank-biased precision. It prices rank \`i\` at \`p^(i-1)\`. Junk near the`,
```
hsjobeki: I would like to move this stuff into an easy to understand footer.
Feeling that this header is hard to understand. A reviewer of a PR needs to understand what the metrics say. That is what is BestRR saying , not what is special about our BestRR, before they know what BestRR is.

I would prefer a footer because it is a lookup place, rather than a wall of text before the relevant numbers.
Compare to the eval-metrics in nixpkgs, where we handle this similarly. 

(just grabbed the report from the last merged PR in nixpkgs for comparison: https://github.com/NixOS/nixpkgs/actions/runs/31798770526/attempts/1#summary-94762578776)

KiaraGrouwstra: hm, these footnotes (updated) maybe look a bit cluttered... 😶

KiaraGrouwstra: wdyt @hsjobeki?

hsjobeki: I like that you don't have to first read through this wall of text that only explains part of the numbers.
Maybe if you don't repeat the footnote; Just when you first introduce the term. i.e. in the first "Overall" table?

hsjobeki (approved):

PR #1525, opened 2026-08-15
Add nixpkgs-multiverse

pinage404:
This project is recent but do something that was really missing with Nix Flakes

PR #1527, opened 2026-08-16
fix(ci): let --save-summary work without --elastic-schema-version

KiaraGrouwstra:
The `Check Flake Groups` job has failed on every `flakes/**.toml` PR since #1521 added `--save-summary $GITHUB_STEP_SUMMARY` to `.github/workflows/check-flake-files.yml`. That job runs `--json` and never `--push`, so it has no schema version to pass, and clap rejects the invocation before any work happens:

```
error: The following required arguments were not provided:
    --elastic-schema-version <elastic-schema-version>
```

Latest failure, on #1525 (whose only change is `flakes/manual.toml`): https://github.com/NixOS/nixos-search/actions/runs/31910644680

`--save-summary` had picked up `requires("elastic-schema-version")` as a copy of the constraint on `--push`, where it is genuine: `push_to_elastic` calls `elastic.elastic_schema_version.unwrap()`. Writing a markdown summary of failures reads no elasticsearch config, so the constraint is dropped here. The one on `--push` stays.

Two tests cover the CLI contract: the workflow's invocation shape parses, and `--push` still demands a schema version. `flake-info/default.nix` runs `cargo test`, so the `Build flake-info` workflow catches a regression.

Follow-up, not addressed here: #1521 notes that `--save-summary` and `--report` overlap and that `--report` may not work.

### DIFF
--- a/flake-info/src/bin/flake-info.rs
+++ b/flake-info/src/bin/flake-info.rs
@@ -36,8 +36,7 @@ struct Args {
 
     #[structopt(
         long = "save-summary",
-        help = "Save markdown summary of failures to this file",
-        requires("elastic-schema-version")
+        help = "Save markdown summary of failures to this file"
     )]
     save_summary: Option<String>,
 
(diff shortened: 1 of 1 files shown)


raboof (approved):
oops, thanks

pinage404:
Thanks !

PR #1542, opened 2026-08-22
Index how popular an option's package is

KiaraGrouwstra:
Give option documents the popularity signal package documents already have, derived from the package their module configures.

A package carries `package_repology_repos`, the number of distribution repositories that ship it, and the query reads it as a `rank_feature` to settle ties between otherwise equally good matches. An option carries nothing comparable: the index does not say that `services.nginx.enable` is wanted more often than `services.lighttpd.enable`, so a query matching both is decided on name length or on Lucene's tie order.

Nixpkgs already draws the link. A module that wraps a package declares which one under `<module path>.package`, with the package as the default:

```
services.nginx.package         default: pkgs.nginxStable
services.postgresql.package    default: pkgs.postgresql_15
```

So an option can be attributed to a package by walking up its own name until a `.package` sibling is found, and inherit that package's count. Modular services name theirs outright through `service_package` and skip the walk.

## Coverage

Against the current nixpkgs index (`nixos-51-unstable-ffb3c9b700e759be2ef13237c9d8f953b32a1e46`, 24897 option documents):

| | |
| --- | --- |
| `*.package` options of `types.package` | 1291 |
| of those, with a `pkgs.<attr>` default | 1271 |
| option documents attributed to a package | 18276 (73.4%) |
| distinct packages they land on | 1226 |
| of those, with a repology count | 1188 |
| option documents that end up with a value | 17858 (71.7%) |

The counts spread widely enough to rank on: min 1, p25 7, median 21, p75 88, p90 123, max 172.

## Two fields

`option_popularity` is the `rank_feature`. `option_package` is the attribute it came from, mapped as a `keyword`, because a `rank_feature` cannot be sorted, aggregated or scripted - without the keyword beside it there is no way to ask the index which package an option was attributed to, and so no way to check the derivation on real data.

## No query uses them yet

A `rank_feature` clause on `option_popularity` was fitted against #1536's benchmark. On the full corpus it looks like a win - nDCG@10 0.717 -> 0.728, Success@10 0.857 -> 0.874 - but on the 70/30 held-out split that #1541's shape was selected under, every pivot and boost tried gains on train and loses on test:

| variant | train fitness | test fitness | test nDCG@10 |
| --- | --- | --- | --- |
| no clause | 0.7271 | 0.7620 | 0.807 |
| `saturation` pivot 50, boost 0.05 | 0.7323 | 0.7566 | 0.799 |
| `saturation` pivot 50, boost 0.2 | 0.7357 | 0.7554 | 0.796 |
| `saturation` pivot 50, boost 0.5 | 0.7421 | 0.7549 | 0.795 |
| `saturation` pivot 21, boost 0.2 | 0.7356 | 0.7574 | 0.799 |
| `saturation` pivot 150, boost 0.3 | 0.7357 | 0.7552 | 0.796 |

The category it was meant to help - `intent`, describing a capability rather than naming a module - stays at 0.114 nDCG@10 in every variant. So the clause is not shipped and only the signal is, for the next run of the query search to consider.

## Deployment

No `version.nix` bump. An import writes a fresh `nixos-<schema>-<channel>-<rev>` index with the current mapping and moves the alias to it, the two fields are additive, and no existing query names them.

Claude-Session: https://claude.ai/code/session_01JfWqryqG18MLe4hP5Se9N1

### DIFF
--- a/flake-info/src/data/export.rs
+++ b/flake-info/src/data/export.rs
@@ -246,6 +246,10 @@ pub enum Derivation {
         option_example: Option<DocValue>,
 
         option_flake: Option<ModulePath>,
+
+        option_package: Option<String>,
+
+        option_popularity: Option<u64>,
     },
     #[serde(rename = "service")]
     Service {
@@ -256,6 +260,8 @@ pub enum Derivation {
         option_default: Option<DocValue>,
         option_example: Option<DocValue>,
         option_flake: Option<ModulePath>,
+        option_package: Option<String>,
+        option_popularity: Option<u64>,
         service_package: Option<String>,
         service_module: Option<String>,
         service_packages: Vec<String>,
@@ -269,6 +275,8 @@ pub enum Derivation {
         option_default: Option<DocValue>,
         option_example: Option<DocValue>,
         option_flake: Option<ModulePath>,
+        option_package: Option<String>,
+        option_popularity: Option<u64>,
     },
     #[serde(rename = "darwin-option")]
     DarwinOption {
@@ -279,6 +287,8 @@ pub enum Derivation {
         option_default: Option<DocValue>,
         option_example: Option<DocValue>,
         option_flake: Option<ModulePath>,
+        option_package: Option<String>,
+        option_popularity: Option<u64>,
     },
 }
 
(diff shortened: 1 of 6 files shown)


KiaraGrouwstra:
hm, given this did not improve results on that query type, i'm not sure if adding it to the index would be beneficial, even if this could still be included in a future evolutionary search

PR #1544, opened 2026-08-25
flakes: add tomodachi94/dotfiles

tomodachi94:
I've got a couple of packages available here.

### DIFF
--- a/flakes/manual.toml
+++ b/flakes/manual.toml
@@ -597,6 +597,11 @@ repo = "kconfq"
 
 [[sources]]
 type = "github"
+owner = "tomodachi94"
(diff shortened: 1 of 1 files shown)


raboof (approved):
not super sure such 'nested' packages are picked up correctly, let's see I suppose :)

tomodachi94:
Should be fine, I expose the packages from both `/flake.nix` and `/pkgs/flake.nix` (the latter only exists so that people don't need to pull in all of my inputs just to use a package).

raboof:
Jep seems like it (e.g.  https://search.nixos.org/flakes?query=hudkit )

PR #1547, opened 2026-08-26
flake-info: index desktop entries and AppStream screenshots

KiaraGrouwstra:
Index each package's desktop entries and AppStream data, so the front-end can use their categories, MIME file types, icons and screenshots (https://github.com/NixOS/nixos-search/pull/1548).

Data is read daily from the binary cache for `nixos-unstable`.

Screenshot are saved as WebP as 224 px wide thumbnails (~5 KB) + 752 px wide larger versions on click  (~29 KB), totaling 1066 packages with 2835 images (15 MB thumbnails + 82 MB main copies).

(NixOS/nixpkgs#556838 extends this with eval-level `desktopEntries`. This is not required: without it the importer indexes whatever the scan found, which should make for the bulk of desktop entries available.)

Assisted-by: Claude:claude-fable-5-1

https://claude.ai/code/session_016cXwBfy83TRabqdcX1UYqb

edit: CI is blocked on #1565.


flake-info/src/data/export.rs
```diff
@@ -218,6 +222,17 @@ pub enum Derivation {
         package_homepage: Vec<String>,
         package_position: Option<String>,
         package_modular_services: Vec<String>,
+        /// The package's desktop entries, for rendering. Kept nested rather than
+        /// flattened: an entry's `type` would otherwise collide with the tag
+        /// this enum is discriminated by.
+        package_desktop_entries: Vec<DesktopEntry>,
+        /// Icons for the entries above, keyed by name, as `data:` URIs. Held
```
KiaraGrouwstra: would this `data:` approach not be less friendly to caching at build and HTTP fetch compared to hosted files (given we expect not too many clashes in terms of application names, and we generally just need to store the newest set)?

flake-info/src/commands/nixpkgs_info.rs
```diff
@@ -1,25 +1,279 @@
...
+}
+
+/// Interns images under content-addressed file names, so that a package
+/// document carries a name an `<img>` can point at rather than the image
+/// itself. Each distinct image is emitted once, as its own document, from
+/// which the frontend build writes a static file.
+struct ImageStore {
+    /// The formats this store may serve, as file extensions.
+    extensions: &'static [&'static str],
+    files: HashMap<String, String>,
```
ysndr: not apparent that this is actually icon-name -> base64(icon)



Haven't checked, yet, but a) what is the "icon name" aand b) whats its scope (store per package, icons-names named by package)


i'm also not quite sure a tool should store files in memory.
as eventually we seem to want to serve ca'ed  paths, would it make more sense to have this abstraction manage _files_?

flake-info/src/commands/nixpkgs_info.rs
```diff
@@ -1,25 +1,279 @@
...
+            }
+        }
+    }
+}
+
+/// Interns images under content-addressed file names, so that a package
+/// document carries a name an `<img>` can point at rather than the image
+/// itself. Each distinct image is emitted once, as its own document, from
+/// which the frontend build writes a static file.
+struct ImageStore {
```
ysndr: id think of moving this into its own module instead of cluttering the command impl.

flake-info/src/commands/nixpkgs_info.rs
```diff
@@ -1,25 +1,279 @@
...
+/// because a screenshot is a photograph of a window rather than a symbol: the
+/// lossy formats are the right ones for it, and vector art never is.
+const SCREENSHOT_EXTENSIONS: [&str; 4] = ["png", "jpg", "jpeg", "webp"];
+
+/// Fills in icons for entries that arrived without one, from a freedesktop icon
+/// theme directory such as `${papirus-icon-theme}/share/icons/Papirus`.
+/// Evaluation knows an entry's icon *name* but never its file, so entries that
+/// did not come through the binary-cache scanner have nothing to render without
+/// a theme to look them up in.
+struct IconTheme {
```
ysndr: this fallback seems rather above and beyond imo. Not useless but kind of blowing up the change.

flake-info/src/commands/nixpkgs_info.rs
```diff
@@ -1,25 +1,279 @@
...
-use crate::data::import::{NixOption, NixpkgsEntry, Package};
+use crate::data::import::{DesktopEntry, NixOption, NixpkgsEntry, Package, Screenshot};
 
 /// Wrapper for the channel `packages.json` format.
 #[derive(Deserialize)]
 struct PackagesInfo {
     packages: HashMap<String, Package>,
 }
 
+/// The output of `flake-info/scripts/desktop-entries-index.py`, which reads
```
ysndr: why does this go through a python script instead of being a native impl?

